PR #3, opened 2024-04-08
Update README.md

Adarshagupta:


miurla (commented):
IMO: Assuming prior knowledge of deployment and considering the presence of the Vercel button, I believe there is no need for changes.

PR #29, opened 2024-04-13
Allow users to customize the model

winniesi:
If not set, the default is to use gpt-4-turbo.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -2,6 +2,10 @@
 # If you need to set a BASE URL, uncomment and set the following:
 # OPENAI_API_BASE=
 
+# Used to set the model for OpenAI API requests.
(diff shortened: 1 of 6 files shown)


miurla (commented):
This is a good pull request. Thank you!

The task manager is also specifying the model, so I'd like you to make a similar change there as well.
Currently, gpt-3.5-turbo is set, but to avoid complexity, let's set gpt-4-turbo as the default.

https://github.com/miurla/morphic/blob/main/lib/agents/task-manager.tsx#L14

`model: openai.chat(process.env.OPENAI_API_MODEL || 'gpt-4-turbo'),`

miurla (approved):
Thank you!

winniesi:
Super happy to see such a great project!

PR #58, opened 2024-04-20
Add writer agent that support specific APIs

miurla:
Add an agent that only output the final answer text, without tools or JSON output.
If set, modify so that the specified agent provides the final answer.
Allow specifying an OpenAI-compatible API via environment variables.

The LLAMA3 model of the Groq API is verified.

Environment variables for specifying the Groq API:
```
USE_SPECIFIC_API_FOR_WRITER=true
SPECIFIC_API_BASE=https://api.groq.com/openai/v1
SPECIFIC_API_KEY=gsk_xxx
SPECIFIC_API_MODEL=llama3-70b-8192
```


miurla:
@gandli Instead of OpenAI, it has become possible to use Gemini. https://github.com/miurla/morphic/pull/191

AIisaweapon:
Not working.

PR #86, opened 2024-04-26
Enable auto expanding textarea

albertdbio:
This replaces the input box with an auto expanding textarea.
However by default when the text area expand, the border radius will eventually cause it to look oval.
We can solve this issue by reducing the border radius each increment.
The key insight is that a border radius of 32 provided the same effect as rounded-full.
Therefore we can decrement it in an algorithmic manner.

Also the behavior of enter is maintained, that is, enter submits while shift+enter adds a new line.

auto-expanding-radius


miurla (changes requested):
Nice work! Good implementation.

Since the initial state design has changed,
could you make the height and padding the same as the current ones?

main | feat
-- | --
Screenshot 2024-04-26 at 13 22 09 | Screenshot 2024-04-26 at 13 21 40

albertdbio:
On it!

albertdbio:
Ahh I see the reason why the pr is 70px.
The auto resize component overrides styles based on font-width and line-height.

Just noticed the previous open issue perhaps the other contributor can give it a try.
Otherwise, I'll try to get it working over the weekend.
However I'm thinking that we either have to accept a font size of 16px or accept a slightly bigger textarea height, possibly 50px.

albertdbio:
@Techno-o76 want to give the requested changes a try?
If it's to hard then perhaps we should create our own auto expanding textarea component that accounts for border radius as that is a very common pattern

miurla:
@albertdbio 
Understood. I'll also try to see if I can resolve it by adjusting the design.

miurla:
Hey @albertdbio,

I've created a v0.1.8 branch, so could you change the base branch to this one?
After merging, I will handle the remaining adjustments on my end.

https://github.com/miurla/morphic/tree/rc-0.1.8

miurla (approved):

albertdbio:
Hey @miurla, I added the bunlock file from rc 0.1.8

miurla:
Thank you!!

albertdbio:
Oh but that also means that react-textarea-autosize is missing in that bun.lockb.
But when you run `bun i`again it will add it

PR #113, opened 2024-05-05
Fix Cloudflare Pages build error.

kalashnikov:
Hi, 

I am trying to deploy it on Cloudflare Pages and I got build error since yesterday. 
I got below error message 

```
The following routes were not configured to run with the Edge Runtime:
    - /_not-found
```

And I found below page might be related and it does fix the error.
https://nextjs.org/docs/app/api-reference/file-conventions/not-found

I am not next.js expert - just tried to deploy it on Cloudflare Pages for use. 
Not sure it is correct way to solve it.

### DIFF
--- a/app/not-found.tsx
+++ b/app/not-found.tsx
@@ -1,0 +1,11 @@
+import Link from 'next/link'
+
+export default function NotFound() {
+  return (
+    <div>
+      <h2>Not Found</h2>
+      <p>Could not find requested resource</p>
+      <Link href="/">Return Home</Link>
+    </div>
+  )
+}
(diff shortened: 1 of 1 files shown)


miurla:
@kalashnikov 
Thank you for the PR. Since the error is resolved, it seems like a good workaround.
However, not-found.tsx is supposed to be prepared only when a custom page is needed, so it shouldn't normally be necessary. No error occurs with Vercel.
Further investigation is needed to determine if this change is necessary. Similar reports have also been made.
https://github.com/cloudflare/next-on-pages/issues/413#issuecomment-1976355283

Therefore, I have created an issue and referred to this PR. #114
Please do not merge this PR, but close it.

kalashnikov:
Got it. Thanks for the feedback. I will close it.

PR #121, opened 2024-05-08
chore panel: better tip for actions display

superlbr:


app/actions.tsx
```diff
@@ -87,7 +87,10 @@ async function submit(formData?: FormData, skip?: boolean) {
   async function processEvents() {
     let action: any = { object: { next: 'proceed' } }
     // If the user skips the task, we proceed to the search
-    if (!skip) action = (await taskManager(messages)) ?? action
+    if (!skip) {
+      uiStream.update(<span className='flex items-center'><Spinner /><span className='ml-2'>Analysis..</span></span>)
```
miurla, line 91: Adding a loader here causes the loader's position to misalign when the collapsible elements are displayed. Therefore, I intentionally did not add it here.

app/actions.tsx
```diff
@@ -117,7 +120,7 @@ async function submit(formData?: FormData, skip?: boolean) {
     let toolOutputs: ToolResultPart[] = []
     let errorOccurred = false
     const streamText = createStreamableValue<string>()
-    uiStream.update(<Spinner />)
+    uiStream.update(<span className='flex items-center'><Spinner /><span className='ml-2'>Generating..</span></span>)
```
miurla, line 123: The loader here is only displayed during tool calls. The message is not appropriate. I don't think there's a need to purposely display a message.

miurla (commented):
Thank you for the PR. 
However, I don't think this change is necessary.

superlbr:
yeah, just friendly for users.

Issue #92, opened 2024-04-28
Enable deep searching of specific URLs


miurla:
When a specific URL is included in the query, retrieve the data from that URL instead of searching.

miurla:
Try this: https://github.com/mendableai/firecrawl

albertdbio:
Firecrawl is pretty slow when I tested it
Is this fine? I'm sure the user would understand as long as we provide a spinner

albertdbio:
Ahh nevermind, the crawl api which scrapes multiple pages is slow, just scraping one page is decent

miurla:
It's a good try!
I got the following advice on our Discord. I'm going to try this.

> You don't need firecrawl for this, IMHO, can just use Playwright/fetch and then use turndown to convert html to markdown

albertdbio:
Firecrawl uses playwright + turndown 😆
We can also use this one https://jina.ai/reader

I'll give it a shot, already got the researcher able to refer to the previous sources it pulled, just that the types are a bit tricky.
Just need to integrate those API

I'll work on this one first and I'll see if I can get back to video section next week, it's been busy at work

albertdbio:
jina seems to me to be a tad bit faster btw, if you wanna try it out

miurla:
Thank you. I was stuck with the chat history feature. I'll be able to start working on a different task soon.

albertdbio:
Hey @miurla, fixed the types related to enabling follow up conversations and also found a way to enable function calling with groq so now I'm moving forward with adding FireCrawl, planning to push a PR by Sunday!

miurla:
@albertdbio Wow, that's great. Is the function calling an implementation that uses groq's sdk to replace the researcher's model? Looking forward to the PR.

miurla:
@albertdbio I've merged significant changes, so it's best to develop from main.

albertdbio:
@miurla not using the groq sdk, it turns out the vercel sdk needed to be upgraded to a minor version. That minor version set's content as optional in the OpenAI zod schema. Groq doesn't return content when doing function calls so it would throw an invalid response error.
Sounds good, I'll use main!

albertdbio:
Got caught up with work over the weekend. Will continue chipping away at it this week. First I'll rebase on to main.

miurla:
I'm working on this feature.

PR #155, opened 2024-05-18
Add Tool to Retrieve Page Content from Specific URLs

miurla:
## Overview
This PR introduces a tool that retrieves content from specific URLs provided by the user. The tool leverages the Exa API to fetch the content and display it in the UI.

## Details
- Added a new tool in `lib/agents/tools/retrieve.tsx` that retrieves content from the web.
- The tool requires an Exa API Key to function.
- If the retrieval is successful, the content is displayed using the `RetrieveSection` component.
- In case of an error, an appropriate message is displayed to the user.

## Requirements
- Exa API Key must be set in the environment variables.

## Closes
- Issue #92

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -32,3 +32,7 @@ UPSTASH_REDIS_REST_TOKEN=[YOUR_UPSTASH_REDIS_REST_TOKEN]
 # enable the share feature
 # If you enable this feature, separate account management implementation is required.
 # ENABLE_SHARE=true
+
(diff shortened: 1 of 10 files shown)


arsaboo:
@miurla why not use https://jina.ai/reader/? Seems like a much better fit.

miurla:
@arsaboo Ah, I hadn't tried jina.ai/reader. It seems to fit. Thank you.

PR #157, opened 2024-05-18
Refactor retrieveTool to use jina.ai

miurla:
#155 

`jina.ai/reader` will be used instead of exa.ai because it offers a larger free tier and the response isn't bad either.
https://github.com/miurla/morphic/pull/155#issuecomment-2118858686


arsaboo:
Can you please indicate how to invoke it? Just including a url will automatically work?

miurla:
As long as you use it within the free tier, there is no need to configure anything.

For more details, please check the documentation.
https://jina.ai/reader/#apiform

arsaboo:
I meant, how do we use it in morphic? Just put the url?

miurla:
Just enter the URL and your question. There's an example on morphic.sh, so give it a try.

https://www.morphic.sh/share/XG6wMdV

PR #164, opened 2024-05-22
prevented search submit when no input value

techcontributor:
returned from the function when there is not input value


miurla (commented):
Nice catch! This issue only occurs on the onKeyDown event, so could you please add a condition to the event to resolve it?

`input.trim().length > 0`


https://github.com/miurla/morphic/pull/164/files#diff-06a509597829dfc3e720c47b51047e08ba858772f51c5dfd2e01bd6deefb4241R115-R119

techcontributor:
@miurla please check and review the recent change

Could you please merge it if everything is going right.

Thank you!

miurla (approved):
Thank you!

PR #172, opened 2024-05-23
removed unused state and logic

techcontributor:
I checked the state `isButtonPressed`, it was never being set to `true`, so the logic is staying unused

If you want to build some logic with the state, let me know


miurla (approved):
There was code left over from a previous version. Thank you!

https://github.com/miurla/morphic/blob/v0.1.7/components/chat-panel.tsx#L55

PR #190, opened 2024-05-31
feat(utils): Add Ollama support and improve error handling for models

peperunas:
This PR introduces support for Ollama and also refactors the existing model initialization to improve readability and maintainability.

---

- Added support for Ollama in `lib/utils/index.ts`.
- Modified getModel function to check whether the necessary environment variables are provided.
- Refactored the code inside getModel for better readability and maintainability.
- If Ollama's environmental variables are not set, we fall back to OpenAI.

PR #200, opened 2024-06-03
draft: Update Ollama support 

miurla:
ref: #190 

- Fix the error: `Unsupported role: tool`
- Use the Ollama sub-model

## The model used
- OLLAMA_MODEL: `mistral`
- OLLAMA_SUB_MODEL: `phi3`


peperunas:
Thank you for your work @miurla! Unfortunately, as you said, this is very unstable and I was not able to get Ollama working.

I tried different models but - one way or the other - I cannot get any result back. I guess we need to way for a fix upstream.

arsaboo:
How are https://github.com/rashadphz/farfalle and https://github.com/ItzCrazyKns/Perplexica able to achieve this? Both support Ollama and work very well.

miurla:
@peperunas Thank you for checking.

>  I guess we need to way for a fix upstream.

Yes, to work stably with Ollama, a fix in the upstream is necessary. Since it cannot be merged into the main, I intend to create an experimental branch so that other developers can test it. I appreciate your work!

miurla:
@arsaboo Those projects are not Generative UI projects.

peperunas:
So, I made a mess and instead of actually pushing here the merges between this branch and `main`, I created another PR 🤦 

Sorry @miurla!

PR #217, opened 2024-06-09
upgrade nextjs 15

superlbr:
here https://nextjs.org/blog/next-15-rc#react-19-rc
fix type error
fix npm package mis in pnpm

### DIFF
--- a/app/actions.tsx
+++ b/app/actions.tsx
@@ -101,7 +101,7 @@ async function submit(
   }
 
   async function processEvents() {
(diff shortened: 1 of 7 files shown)


miurla:
@superlbr 

Nextjs15 is still in the RC stage, and there are no significant feature additions that would greatly benefit morphic, so we want to proceed with the update cautiously. For now, I'll close this PR. Thank you for your contribution.

PR #228, opened 2024-06-15
fix: The fix for image loading error handling in video search results…

Dogtiti:
… may cause an infinite loop.


miurla (commented):
I don't think this fix is necessary. The placeholder image URL is a static resource and is hardcoded. It's rare for this URL to be undefined or invalid.

PR #242, opened 2024-06-22
feat: Dynamic example messages in empty-screen component 

nohablojavanese:
### Overview
Enhances the EmptyScreen component by introducing simple dynamic message rotation using React's state (useState) and lifecycle method (useEffect). It allows for displaying multiple messages and cycling through them automatically over time, providing a more interactive and dynamic user experience compared to the static example in the initial version.


miurla (commented):
Thank you for the PR! However, it seems that the changes have disappeared in this commit. Please check: https://github.com/miurla/morphic/pull/242/commits/b118f7343b130106f1b82ab4178a6c8c6b465d0e

PR #288, opened 2024-08-01
fix of issue: [BUG] groq cannot be used

calathus:
I noticed, morphic use function call, so model like llama-3.1-70b-versatile which does not support it will fail. So right now there only two such models in groq(8/1/2024). llama3-groq-70b-8192-tool-use-preview, llama3-groq-8b-8192-tool-use-preview.
-- but if we don't need such feature, it might be better to use normal models. so there should be such internal optimization to switch model .
groq is quit popular, it is better to use dedicated variables like GROQ_API_KEY rather then using generic variable.

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -19,16 +19,33 @@ export function getModel(useSubModel = false) {
   let openaiApiModel = process.env.OPENAI_API_MODEL || 'gpt-4o'
   const googleApiKey = process.env.GOOGLE_GENERATIVE_AI_API_KEY
   const anthropicApiKey = process.env.ANTHROPIC_API_KEY
+  // [nc]
+  const specificApiKey = process.env.USE_SPECIFIC_API_FOR_WRITER && process.env.SPECIFIC_API_KEY
 
   if (
     !(ollamaBaseUrl && ollamaModel) &&
     !openaiApiKey &&
     !googleApiKey &&
-    !anthropicApiKey
+    !anthropicApiKey &&
+    !specificApiKey // [nc]
   ) {
     throw new Error(
       'Missing environment variables for Ollama, OpenAI, Google or Anthropic'
     )
+  }
+  // [nc] mainly for groq
+  if (specificApiKey) {
+    const specificApiBase = process.env.SPECIFIC_API_BASE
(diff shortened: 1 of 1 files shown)


miurla:
Thank you for the pull request. I deployed it to Vercel, but I will respond without using the tool. Is it working correctly on your local environment?

## Screenshot
image

## Env
image

miurla:
I tried all the Morphic agents in the local environment using the Groq tool model, but they didn't work due to errors. I don't think it's possible to fully operate Morphic with an OpenAI-compatible model.

```
Error: Model does not have a default object generation mode.
    at generateObject (/Users/yoshiki/Works/next/morphic/.next/server/chunks/ssr/_9c4c19._.js:756:23)
```

miurla:
@calathus  I checked it again. Please check the following.

## Check list
1. The environment variables are set correctly. The Groq API is being called: https://github.com/miurla/morphic/pull/288#issuecomment-2264786669
2. When checking the preview of your PR, it responds without searching for any prompt. The Groq model does not call the function. You can also confirm this in the preview: https://morphic-git-fork-calathus-groq-suport-morphic-ai.vercel.app/
3. I noticed that Groq has been added to the Vercel AI SDK Provider. The model list includes the latest `llama-3.1-70b-versatile`, which also supports tool streaming: https://sdk.vercel.ai/providers/ai-sdk-providers/groq
4. Changing the Groq model to `llama-3.1-70b-versatile` calls the function. However, I tried several models, but errors occurred and it could not operate stably.

## Conclusion
Currently, Morphic cannot operate stably with only the Groq model. We will support the Groq provider when it is updated to a stable model. Therefore, it is optimal to apply the Groq API only for text generation as in the current code. Also, since Groq is not the only OpenAI-compatible API, it is correct to provide it as a generic environment variables.

miurla:
I have created an issue for the Groq Provider: #289 

Please close this PR once you have confirmed it. I appreciate your work.

miurla:
> but the instability may be due to the rate  limit, and the paid users may not have much problem.

No, it's not an error related to rate limits; issues such as JSON errors or incomplete responses occur. These problems need to be resolved.

`Tool use failed: JSON does not match the expected schema for tool calls`


Additionally, when addressing these issues, it's necessary to not only initialize the model but also modify the main logic. So I will close this PR.

PR #317, opened 2024-08-21
Update and rename .env.local.example to .env

gionmud0:


miurla:
@gionmud0 Your API key has been exposed. You should disable the key immediately.

PR #322, opened 2024-08-25
Fix searxng settings

tidymonkey81:


tidymonkey81:
oops. no idea what happened here. appologies.

PR #345, opened 2024-09-24
Date, missing almost versatile UI elements

gionmud0:
Hey, if you have some time, please fix, I'd provide issue template, but it's impossible, haven't changed either env. Project deployed with a fork of repo, forks of origin couldn't find. Thanks.


miurla:
Your API Key has become public. You should invalidate it immediately.

PR #355, opened 2024-09-30
feat: Support Groq AI Provider

miurla:
Until now, Groq was only applied to the Writer agent, but it has been made selectable as a provider option.
Only models compatible with tool use are available.
Set the following environment variables:

```
# GROQ_API_KEY=[YOUR_GROQ_API_KEY]
# GROQ_API_MODEL=[YOUR_GROQ_API_MODEL] # e.g. llama3-groq-8b-8192-tool-use-preview
```

Along with this change, the writer agent and `SPECIFIC_API_xxx` have been removed.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -57,6 +57,10 @@ SEARXNG_SAFESEARCH=0 # Safe search setting: 0 (off), 1 (moderate), 2 (strict)
 # If you want to use Anthropic instead of OpenAI, enable the following settings.
 # ANTHROPIC_API_KEY=[YOUR_ANTHROPIC_API_KEY]
 
+# If you want to use Groq, enable the following variables. only available for tool use compatible models.
+# GROQ_API_KEY=[YOUR_GROQ_API_KEY]
+# GROQ_API_MODEL=[YOUR_GROQ_API_MODEL] # e.g. llama3-groq-8b-8192-tool-use-preview
+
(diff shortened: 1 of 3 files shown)


freedomtrain:
Thanks for all of you work !!!
If we use this in env.local do we also need to continue to use the openAI key and ChatGPT ?
# GROQ_API_KEY=[YOUR_GROQ_API_KEY]
# GROQ_API_MODEL=[YOUR_GROQ_API_MODEL] # e.g. llama3-groq-8b-8192-tool-use-preview
Can you please also please help with an env.local example of what needs to be added to have groq be used instead of OpenAI

miurla:
> If we use this in env.local do we also need to continue to use the openAI key and ChatGPT ?

The latest version allows you to select models in the UI: https://github.com/miurla/morphic/pull/371. You only need to set the keys for the providers you need. If you don't need OpenAI, you don't need to set it.

> Can you please also please help with an env.local example of what needs to be added to have groq be used instead of OpenAI

Just set the API key: https://github.com/miurla/morphic/blob/main/.env.local.example#L48

If you want to add another model, please edit this file: https://github.com/miurla/morphic/blob/0a0498e290dfa19352418ee3e5a33a1107eb6101/lib/types/models.ts#L39C1-L44C5

Only models compatible with tool use are available.

Issue #347, opened 2024-09-29
Add No-History Options


miurla:
### Feature Description

Related: #337 

Enable building without Redis setup when history is not needed.

### Use Case

In case you want to simply use only the search function.

### Additional context

We can make the Redis setup optional, allowing for easier building.

miurla:
Made it configurable through environment variables
Can be enabled with `NEXT_PUBLIC_ENABLE_SAVE_CHAT_HISTORY=true`

https://github.com/miurla/morphic/blob/main/.env.local.example#L21

ref: #399

PR #369, opened 2024-10-23
Toggle history

casistack:
## Detailed Changes

1. **Chat History Toggle Feature**:
   - Implemented user-controlled chat history toggle
   - Added settings UI for enabling/disabling chat history
   - Real-time UI updates for history panel
   - Proper state management for history visibility

2. **Component Updates**:
   - Enhanced `HistoryList` component to use client-side state management for immediate UI updates
   - Updated `ClearHistory` component to properly refresh the history list after clearing
   - Modified `History` component to handle state transitions and loading states
   - Added proper handling for history refresh operations

3. **Redis Integration**:
   - Modified the Redis implemetation to support credentials
   - Implemented error handling for Redis operations
   - Added fallback behavior when Redis operations fail   

4. **Storage Implementation**:
   - Added flexible storage provider system
   - Implemented Redis storage adapter for server-side storage
   - Added browser-based storage (localStorage) for development/fallback
   - Created storage abstraction layer for future extensibility

5. **Environment Variables Added**:
   - `STORAGE_PROVIDER`: Choose between 'redis' or 'local' storage
   - Enhanced Redis configuration options
   - Added fallback behavior configuration

## Testing Instructions

1. **Chat History Toggle**:
   - Enable/disable chat history in settings
   - Verify chat persistence matches setting
   - Test real-time updates in history panel
   - Verify clear history functionality

2. **Storage Testing**:
   - Test with Redis:     ```env
     STORAGE_PROVIDER=redis
     USE_LOCAL_REDIS=false
     UPSTASH_REDIS_REST_URL=[URL]
     UPSTASH_REDIS_REST_TOKEN=[TOKEN]     ```
   - Test with LocalStorage:     ```env
     STORAGE_PROVIDER=local     ```
   - Test fallback behavior with invalid Redis config

## Potential Impacts

- Improved user privacy control with history toggle
- Enhanced development experience with LocalStorage option
- No breaking changes to existing functionality
- Seamless fallback behavior for reliability

## Additional Notes

- Storage abstraction allows for future storage providers
- Browser-based storage (localStorage) makes development and testing easier
- Data in browser storage is device-specific and persists until cleared
- All changes maintain backward compatibility
- Added proper error handling and state management
- UI updates happen in real-time without page refreshes

## Checklist Additions

- [x] Implemented chat history toggle feature
- [x] Added storage provider abstraction
- [x] Implemented LocalStorage provider
- [x] Added proper error handling
- [x] Updated documentation
- [x] Tested all storage scenarios
- [x] Verified backward compatibility
- [x] Added some TypeScript types for changes
- [x] Implemented real-time UI updates

Closes #[347] Add No-History Options


miurla:
@casistack This is another great update! Thank you! 🥇 

Please fix the compilation error that is occurring.
```
[09:38:56.714] Running "bun run build"
[09:38:56.719] $ next build
[09:38:57.382]   ▲ Next.js 14.2.5
[09:38:57.382] 
[09:38:57.457]    Creating an optimized production build ...
[09:39:26.506]  ✓ Compiled successfully
[09:39:26.507]    Linting and checking validity of types ...
[09:39:34.144] Failed to compile.
[09:39:34.144] 
[09:39:34.144] ./components/history-container.tsx:17:10
[09:39:34.144] Type error: Property 'chatHistoryEnabled' is missing in type '{ userId: string; }' but required in type 'HistoryListProps'.
```

components/history-list.tsx
```diff
@@ -30,7 +61,7 @@ export async function HistoryList({ userId }: HistoryListProps) {
         )}
       </div>
       <div className="mt-auto">
-        <ClearHistory empty={!chats?.length} />
+        <ClearHistory empty={!chats?.length} onCleared={refreshList} />
```
miurla: The Clear button position is shifted downward

image

components/header.tsx
```diff
@@ -16,6 +17,7 @@ export const Header: React.FC = async () => {
       <div className="flex gap-0.5">
         <ModeToggle />
         <HistoryContainer location="header" />
+        <TopRightMenu />
```
miurla: Thank you for adding the menu.

The ChatHistoryToggle appears in two places - in the history and menu sections, but I think one location would be sufficient. Since ChatHistoryToggle is currently the only active menu item, should we keep the header menu hidden for now?

components/history-list.tsx
casistack: oops ok done

components/header.tsx
casistack: commented it out

app/actions.tsx
```diff
@@ -70,8 +76,16 @@ async function submit(
...
+  // Always add the user message to the messages array if content is not null
+  if (content !== null) {
+    messages.push({
+      role: 'user',
+      content
+    })
+  }
+
+  // Only update aiState if chat history is enabled and content is not null
+  if (chatHistoryEnabled !== 'false' && content !== null) {
```
miurla: `chatHistoryEnabled !== 'false'`

AIState needs to be updated to continue chatting. This change is not necessary.

app/actions.tsx
```diff
@@ -147,8 +157,16 @@ export const AI = createAI<AIState, UIState>({
   onSetAIState: async ({ state, done }) => {
     'use server'
 
+    const redis = await getRedisClient()
+    const chatHistoryEnabled = await redis.get(
```
miurla: I tested with Upstash, and even when disabled, it was always saved. The reason was that `chatHistoryEnabled` was returning a boolean value. Please check this.

```
console.log({
  value: chatHistoryEnabled,
  type: typeof chatHistoryEnabled,
  comparison: chatHistoryEnabled === 'false',
  strictEquality: Object.is(chatHistoryEnabled, 'false')
})
```
```
{
  value: false,
  type: 'boolean',
  comparison: false,
  strictEquality: false
}
```

lib/storage/local-storage.ts
```diff
@@ -0,0 +1,70 @@
+import { StorageProvider } from './types'
+
+export class LocalStorageProvider implements StorageProvider {
+  private getFullKey(key: string): string {
+    return `morphic:${key}`
+  }
+
+  async get(key: string): Promise<string | null> {
+    if (typeof window === 'undefined') return null
```
miurla, line 9: The storage provider is always called in server actions. Therefore, LocalStorageProvider should always return null and shouldn't work.

miurla (commented):
> Test with LocalStorage: `env STORAGE_PROVIDER=local`

LocalStorage wasn't working. To make LocalStorageProvider work, unlike Redis, we need to add logic to save on the client side. ref: https://github.com/miurla/morphic/pull/369#discussion_r1815938400

Since we don't need to go that far, how about making `STORAGE_PROVIDER` options either `redis` or `none`? When set to `none`, nothing will be saved to storage. This would satisfy the requirements of #347.

app/actions.tsx
casistack: hmm thats odd  i tested with local redis and was working  and upstash . ok will have a look

components/header.tsx
```diff
@@ -6,6 +6,9 @@ import HistoryContainer from './history-container'
 import TopRightMenu from './ui/top-right-menu'
 
 export const Header: React.FC = async () => {
+  // Get storage provider setting from environment
+  const storageProvider = process.env.STORAGE_PROVIDER || 'redis'
```
miurla, line 10: The header will only be controlled on mobile. If you return null in the history-container, you can control everything at once.

https://github.com/miurla/morphic/blob/main/components/history-container.tsx
```
  const storageProvider = process.env.STORAGE_PROVIDER
  const enabled = storageProvider !== 'none'

  if (!enabled) {
    return null
  }
```

lib/redis/config.ts
```diff
@@ -115,21 +169,34 @@ export class RedisWrapper {
...
-      return (this.client as RedisClientType).get(key)
+    
+    if (this.shouldSkipOperation()) {
+      return null
+    }
+
+    try {
+      if (this.client instanceof Redis) {
+        const value = await this.client.get(key)
+        return value ? String(value) : null
```
miurla, line 182: I performed verification. In the case of Upstash, the value of `user:${userId}:chatHistoryEnabled` was being returned as a boolean, even though it was stored as a string.
Therefore, when the value was` false`, it was returning either `'true'` or `null`. So, modifying it as follows achieved the expected result.

`return value === null ? null : String(value)`

app/actions.tsx
```diff
@@ -157,19 +148,16 @@ export const AI = createAI<AIState, UIState>({
   onSetAIState: async ({ state, done }) => {
     'use server'
 
+    // Get chat history setting first
     const redis = await getRedisClient()
-    const chatHistoryEnabled = await redis.get(
-      'user:anonymous:chatHistoryEnabled'
-    )
+    const chatHistoryEnabled = await redis.get('user:anonymous:chatHistoryEnabled')
```
miurla, line 166: IMO: Since chatHistoryEnabled is also being checked in saveChat, checking it in either place should be sufficient.

https://github.com/miurla/morphic/pull/369/commits/2804b5e15c5a9f4ea0975eb63dba19419f738883#diff-819d2d9016d052cd621d2190deaf60688a87f953233d709b40fdabcf99d83f47R175-R180

lib/actions/chat.ts
```diff
@@ -17,7 +17,9 @@ export async function getChats(userId?: string | null) {
 
   try {
     const redis = await getRedis()
-    const chatHistoryEnabled = await redis.get(`user:${userId}:chatHistoryEnabled`)
+    const chatHistoryEnabled = await redis.get(
```
miurla, line 20: You can use `getChatHistorySetting`

`const chatHistoryEnabled = await getChatHistorySetting(userId)`

https://github.com/miurla/morphic/pull/369/commits/2804b5e15c5a9f4ea0975eb63dba19419f738883#diff-819d2d9016d052cd621d2190deaf60688a87f953233d709b40fdabcf99d83f47R217

components/chat-panel.tsx
```diff
@@ -149,7 +148,7 @@ export function ChatPanel({ messages, query }: ChatPanelProps) {
             placeholder="Ask a question..."
             spellCheck={false}
             value={input}
-            className="resize-none w-full min-h-12 rounded-fill bg-muted border border-input pl-4 pr-10 pt-3 pb-1 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50'"
+            className="resize-none w-full min-h-12 rounded-fill bg-muted border border-input pl-4 pr-10 pt-3 pb-1 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50'"
```
miurla: The ring color has changed.  I don't want to change the ring color.

image

miurla (commented):
Thank you for fixing the `STORAGE_PROVIDER` 👍  I've made several comments.

app/actions.tsx
miurla: I performed additional testing: https://github.com/miurla/morphic/pull/369#discussion_r1817794027

lib/redis/config.ts
casistack: ah yes . makes sense should have spotted that

app/actions.tsx
casistack: yea just put it as additional safety net but see what you mean

components/chat-panel.tsx
casistack: ok so no cosmetic changes . cool

lib/storage/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    return this.redis.hmset(key, value)
+  }
+}
+
+export async function getStorageProvider(): Promise<StorageProvider> {
+  if (storageProvider) return storageProvider
+
+  const preferredStorage = process.env.STORAGE_PROVIDER?.toLowerCase() || 'redis'
+
+  if (preferredStorage === 'local') {
```
miurla, line 54: I overlooked it. Let's remove the local storage-related code.

- lib/storage/local-storage.ts

miurla:
Along with the migration, I made it possible to enable/disable history using environment variables. I will close this PR 🙏  FYI: @casistack

casistack:
> Along with the migration, I made it possible to enable/disable history using environment variables. I will close this PR 🙏 FYI: @casistack

sorry been busy with work glad you have implemented this

Issue #432, opened 2025-02-14
[BUG] Unable to run the Docker image


davidtavarez:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Vercel Runtime Logs

- [ ] I have checked the Vercel Runtime Logs for errors (if applicable)

### Current Behavior

While using Docker I am using the next error:

```
$ next dev --turbo -H 0.0.0.0
18586 | function load(scopedContext = {}) {
18587 |   const context = {};
18588 |   Object.assign(context, {
18589 |     TextDecoder,
18590 |     TextEncoder,
18591 |     TextEncoderStream,
            ^
ReferenceError: Can't find variable: TextEncoderStream
      at load (/app/node_modules/next/dist/compiled/@edge-runtime/primitives/load.js:18591:5)
      at /app/node_modules/next/dist/compiled/@edge-runtime/primitives/index.js:5:22
      at /app/node_modules/next/dist/server/require-hook.js:65:12
      at /app/node_modules/next/dist/compiled/@edge-runtime/ponyfill/index.js:2:42
      at /app/node_modules/next/dist/server/require-hook.js:65:12
      at /app/node_modules/next/dist/telemetry/storage.js:20:7
      at /app/node_modules/next/dist/server/require-hook.js:65:12
error: script "dev" exited with code 1
```

### Expected Behavior

It should run without problems.

### Steps To Reproduce

1. Create a docker-compose.yml file with the next content:
```yaml
services:
  morphic:
      image: ghcr.io/miurla/morphic:latest
      container_name: morphic
      hostname: morphic
      environment:
        - OLLAMA_BASE_URL=${OLLAMA_BASE_URL}
        - SEARCH_API=searxng
        - SEARXNG_API_URL=http://searxng:8080
        - NEXT_PUBLIC_ENABLE_SAVE_CHAT_HISTORY=true
        - USE_LOCAL_REDIS=true
        - LOCAL_REDIS_URL=redis://redis:6379
        - SERPER_API_KEY=${SERPER_API_KEY}
      ports:
        - '3000:3000'
```
2. Run the next command:
```
docker compose up -d
```

### Environment

```markdown
- OS: Linux
```

miurla:
The fix has been merged and the new Docker image is now available.

Please pull the latest image:

```bash
docker pull ghcr.io/miurla/morphic:latest
```

The image now uses production mode by default, so you don't need any special configuration in your docker-compose.yaml. Just run:

```bash
docker compose up -d
```

[...]

PR #433, opened 2025-02-14
fix: use bun start for production mode in Docker

miurla:
fix: #432 

- Update docker-compose.yaml to use bun start
- Ensure consistency with Dockerfile CMD
- Optimize for production environment


davidtavarez:
Thank you for the fix. I can confirm that the image can be built. Now the same error is display in the UI. https://github.com/miurla/morphic/issues/435

PR #473, opened 2025-03-13
fix: use parts to render  messages in multi steps correctly

sp88011:


miurla:
I noticed that the deployment was failing because `bun update ai` wasn't run. I've created PR #474 which includes your fix along with error corrections and a refactoring of the reasoning component.

We'll merge your PR as is. Thanks for your contribution!

PR #517, opened 2025-05-07
feat: add I18n next-intl

oussamadhouib:


miurla:
Hi @oussamadhouib,

Thanks for your work on introducing internationalization!

We've reviewed the basic setup for `next-intl`, and considering Morphic's specific features, we've identified a few additional points that need further consideration.

Morphic has a sharing feature, and including the locale in the URL path, as in the current implementation, could affect the user experience – for example, by fixing the language of shared URLs. The URL structure needs careful consideration.

Additionally, to actually implement multi-language support, we'll need to implement a language switcher UI and define a translation strategy, including which languages to support and the scope of translation.

Considering these points, we feel it would be best to allow for more preparation time for the internationalization feature and to plan it more thoroughly.
While we appreciate this initial implementation in this PR, we'd like to close this one for now and plan to tackle this feature as a separate version with more comprehensive planning.

Thank you for your efforts!

PR #541, opened 2025-05-16
A middleware-based approach for base URL determination

robertDouglass:
Make baseUrl dynamic.

Updated middleware.ts to:

- Extract protocol and host from request headers
- Construct a dynamic base URL
- Add this information to response headers
- Preserve existing Supabase session functionality

Modified lib/config/models.ts to:

- Export the getBaseUrlFromHeaders() function
- Check for NEXT_PUBLIC_BASE_URL first as an override
- Fall back to dynamically determined base URL from headers

Updated lib/tools/search.ts to:

- Import and use the getBaseUrlFromHeaders() function
- Dynamically determine the base URL when NEXT_PUBLIC_BASE_URL isn't set

The development server can now run on any port (eg. 3001), confirming that the system can automatically adapt to different ports without manual configuration. This makes the application more flexible when deployed in various environments.

### DIFF
--- a/lib/config/models.ts
+++ b/lib/config/models.ts
@@ -17,8 +17,8 @@ export function validateModel(model: any): model is Model {
 
 export async function getModels(): Promise<Model[]> {
   try {
-    // Check for BASE_URL environment variable first
-    const baseUrlEnv = process.env.BASE_URL
+    // Check for NEXT_PUBLIC_BASE_URL environment variable first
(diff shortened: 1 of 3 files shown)


miurla:
Thanks for this PR, @robertDouglass! The dynamic base URL approach works well and handles different deployment scenarios nicely.

Merging this. Thanks for your contribution!

PR #543, opened 2025-05-18
Fix React Hook Warnings

robertDouglass:
React Hook Warnings Fixed:

Removed invalid dependency 'scrollContainerRef.current' from useEffect
Added missing dependencies 'savedMessages' and 'setMessages' to useEffect
Eliminated build warnings for proper React Hook usage

### DIFF
--- a/lib/config/models.ts
+++ b/lib/config/models.ts
@@ -17,8 +17,8 @@ export function validateModel(model: any): model is Model {
 
 export async function getModels(): Promise<Model[]> {
   try {
-    // Check for BASE_URL environment variable first
-    const baseUrlEnv = process.env.BASE_URL
+    // Check for NEXT_PUBLIC_BASE_URL environment variable first
+    const baseUrlEnv = process.env.NEXT_PUBLIC_BASE_URL || process.env.BASE_URL
     let baseUrlObj: URL
 
     if (baseUrlEnv) {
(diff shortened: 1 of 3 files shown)


components/chat.tsx
```diff
@@ -114,7 +114,7 @@ export function Chat({
     handleScroll() // Set initial state
 
     return () => container.removeEventListener('scroll', handleScroll)
-  }, [scrollContainerRef.current])
+  }, [])
```
miurla: 👍

components/chat.tsx
```diff
@@ -133,7 +133,7 @@ export function Chat({
 
   useEffect(() => {
     setMessages(savedMessages)
-  }, [id])
+  }, [id, savedMessages, setMessages])
```
miurla: For this change, I suggest an alternative approach:

If the goal is to fix the ESLint warning while maintaining the original behavior (only running when `id` changes), the best solution would be to use:

```typescript
// eslint-disable-next-line react-hooks/exhaustive-deps
}, [id])
```

This explicitly tells ESLint that we intentionally omitted dependencies because:
- We only want this effect to run when `id` changes
- Adding `savedMessages` and `setMessages` would change the intended behavior
- The original implementation with just `[id]` correctly captures our requirements

robertDouglass:
New PR https://github.com/miurla/morphic/pull/553

Issue #536, opened 2025-05-15
Implement dynamic contextual OG Images for share links


miurla:
## Description
Currently, we have a static OG Image for shared links. We should enhance the social sharing experience by implementing dynamic, contextual OG Images based on the content being shared (search results, chat contexts, etc.).

## Proposed Implementation
- Utilize Next.js's ImageResponse API to generate dynamic OG Images
- Create route-based dynamic images for:
  - Search results (`app/search/[id]/opengraph-image.tsx`)
- Include relevant thumbnails or content previews in the generated images
- Display key information (search terms, chat topics) 
- Maintain brand consistency with fallback to static image when dynamic content isn't available

## Expected Benefits
- Improved social sharing experience
- Higher engagement rates when links are shared on social platforms
- Better visual context for shared links
- Increased click-through rates

## Technical Requirements
- Use Edge runtime for optimal performance
- Implement responsive image generation to accommodate different platforms
- Ensure proper caching strategy
- Maintain accessibility with appropriate alt text

PR #567, opened 2025-06-16
feat: Use image from search results for OG tag

ngoiyaeric:
This corrects the implementation for dynamic Open Graph (OG) images for your search result pages (`/search/[id]`) as per issue #536 and your subsequent feedback.

The `og:image` meta tag will now point to an actual image URL retrieved from the search results, rather than generating a new image card.

Key changes:
- I modified the `generateMetadata` function in `app/search/[id]/page.tsx`:
  - It now fetches chat data, looking for a 'data' message containing search results (images and web results).
  - It extracts the first image URL from the `images` array within the search results.
  - This URL is set as the `og:image` in the page's metadata, including standard width (1200) and height (630).
  - If no image is found in the search data or an error occurs, the `og:image` tag is not set by this function, allowing Next.js to fall back to the global `app/opengraph-image.png`.
- I removed the `app/search/[id]/opengraph-image.tsx` file, as it was based on the previous approach of generating a new image.

Manual testing is required to confirm that:
- Searches with images use an image from the results as `og:image`.
- Searches without images, or invalid searches, fall back to the global `app/opengraph-image.png`.

### DIFF
--- a/app/search/[id]/page.tsx
+++ b/app/search/[id]/page.tsx
@@ -4,7 +4,7 @@ import { getCurrentUserId } from '@/lib/auth/get-current-user'
 import { getModels } from '@/lib/config/models'
 import { convertToUIMessages } from '@/lib/utils'
 import { notFound, redirect } from 'next/navigation'
-import { ExtendedCoreMessage, SearchResults, SearchResultImage } from '@/lib/types'; // Added SearchResults and SearchResultImage
+import { ExtendedCoreMessage, SearchResults } from '@/lib/types'; // Added SearchResults
 
 export const maxDuration = 60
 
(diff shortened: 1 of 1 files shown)


miurla (commented):
Thanks @ngoiyaeric for this clean implementation!

The approach looks solid:
- ✅ Proper handling of both image formats (string/object)
- ✅ Good error handling with fallback to global OG image
- ✅ No performance impact (reuses existing chat data)
- ✅ Follows existing codebase patterns

**Minor suggestion**: As [Copilot mentioned](https://github.com/miurla/morphic/pull/567#discussion_r2151128024), remove the unused `SearchResultImage` import to keep imports clean.

miurla (approved):

PR #597, opened 2025-07-29
feat: Auto-collapse multiple tools in messages

miurla:
## Summary

This PR implements automatic collapsing of tool sections when multiple tools are present in a single message, improving the readability of complex AI responses.

## Changes

- Modified `getIsOpen` function in `chat-messages.tsx` to detect when a message contains multiple tools
- When a message has more than one tool (search, fetch, askQuestion, relatedQuestions), they default to collapsed state
- Single tools maintain the current behavior (open by default)
- User preferences are preserved - manual open/close actions still take precedence

## Behavior

**Before**: All tools in a message are expanded by default, which can be overwhelming when multiple tools are used
**After**: 
- Single tool: Expanded by default (no change)
- Multiple tools: Collapsed by default for cleaner UI
- User can still manually expand/collapse as needed

## Testing

- ✅ TypeScript type checking passes
- ✅ ESLint linting passes
- ✅ Prettier formatting applied
- ✅ Production build successful

Closes TODO item: "tool が複数ある場合は、collasibleを閉じる"

### DIFF
--- a/components/chat-messages.tsx
+++ b/components/chat-messages.tsx
@@ -53,10 +53,35 @@ export function ChatMessages({
     sections.length > 0 &&
     sections[sections.length - 1].assistantMessages.length === 0
 
-  const getIsOpen = (id: string, partType?: string, hasNextPart?: boolean) => {
+  const getIsOpen = (
+    id: string,
+    partType?: string,
+    hasNextPart?: boolean,
+    message?: UIMessage
+  ) => {
     // If user has explicitly modified this state, use that
     if (userModifiedStates.hasOwnProperty(id)) {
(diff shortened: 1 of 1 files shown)


miurla:
Thanks for the review @Copilot! I've addressed both concerns:

1. **Performance optimization**: Replaced the repeated filtering with a Map-based cache that stores tool counts per message ID. This ensures the calculation only happens once per message.

2. **Type safety**: Fixed the `partType || ''` issue by explicitly checking if `partType` exists before using it in the `includes()` check.

The updated implementation is now more performant and type-safe. All tests pass (lint, typecheck, format).

PR #615, opened 2025-08-03
Revert "Fix Anthropic model message structure for tool-call/tool-result separation"

miurla:
Revert: #614


miurla:
## Context on the Revert and Resolution

This PR reverted #614 because while it attempted to fix the Anthropic "tool_use ids were found without tool_result blocks immediately after" error, it was solving the wrong problem.

### Root Cause Analysis

The actual issue was that **`step-start` parts were not being persisted to the database**. In the AI SDK's `convertToModelMessages` function, `step-start` parts trigger message splitting:

```typescript
} else if (part.type === 'step-start') {
  processBlock(); // This creates a new assistant message
}
```

Without `step-start` parts, all assistant message parts were combined into a single message, causing Anthropic's API to reject the message structure.

### Solution in PR #616

PR #616 properly fixed this by:
1. **Persisting `step-start` parts** to maintain message boundaries
2. **Preserving state-based tool structure** for tool-search, tool-fetch, and tool-question (as done in the Vercel reference implementation)

This ensures that messages loaded from the database maintain the same structure as during streaming, preventing the Anthropic API error without requiring any model-specific message conversion.

The fix follows the official Vercel AI SDK persistence pattern: https://github.com/vercel-labs/ai-sdk-persistence-db/blob/main/lib/utils/message-mapping.ts

PR #708, opened 2025-09-23
feat: integrate Ollama model discovery with tools filtering

dan-and:
Add dynamic Ollama model discovery that fetches models from running Ollama instance and filters them based on "tools" capability for websearch-capabilities.. 
This includes new API endpoints and extends model interface

**In detail:** 
- Add OllamaClient for API integration with /api/tags and /api/show endpoints
- Implement dynamic model discovery with tools capability detection
- Add /api/ollama/models and /api/config/models endpoints
- Extend Model interface with capabilities and contextWindow fields
- Update registry to support Ollama tool calling
- Filter models to only include those with tools capability
- Merge static and Ollama models in unified endpoint"

**Screenshot of using morphic with a Ollama Server with several models:** 
 
morphic_ollama_model_discovery

(Models without "tools" capabilities are automatically stripped from the discovery as they can't be used for websearch)

**What may be missing:** 
The static examples (DeepSeek R1 and <OLLAMA MODEL_NAME>) for Ollama could be removed after merging this code, as they are obsolete. I kept that to show the successful combination of the static models and the dynamic discovery.

### DIFF
--- a/app/api/config/models/route.ts
+++ b/app/api/config/models/route.ts
@@ -5,17 +5,20 @@ import { getModels } from '@/lib/config/models'
 export async function GET() {
   try {
     const models = await getModels()
-    
-    return NextResponse.json({ models }, {
-      headers: {
-        'Cache-Control': 'public, max-age=60, s-maxage=60', // Cache for 1 minute
-        'Content-Type': 'application/json'
+
+    return NextResponse.json(
+      { models },
+      {
+        headers: {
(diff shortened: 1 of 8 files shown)


miurla (approved):
Thank you for pushing this Ollama discovery work! CI is only red because bun format needs to run; once that's sorted we're good to merge. After we land this I’ll follow up to remove the static Ollama examples since they’re no longer needed.

miurla:
@dan-and Thank you!

PR #777, opened 2026-02-19
This looks cool

Aaron124us:
i'd like to try it


miurla:
Your Tavily API key and Supabase credentials are exposed in this diff. Revoke them immediately.

This is local config, not a valid contribution. Please don't submit PRs like this (including #776). Closing.

PR #821, opened 2026-04-13
feat: add Gravatar support for user avatars

### LINKED ISSUE #820: [Feature Request] Add Gravatar support for user avatars
### Feature Description

Currently, users who sign up via email and password do not have a profile picture, leading to a generic placeholder. Integrating Gravatar support will automatically fetch user avatars based on their hashed email address. 

### Use Case

Instead of seeing a cold, generic icon, users are immediately greeted with their own familiar profile picture the moment they log in. This makes the application feel much more personalized right from the start. It also brings the Email/Password login experience on par with Google/GitHub OAuth logins which already provide avatars.

### Additional context

This change is non-breaking and only affects the client-side display logic. It significantly improves the UI for the growing number of users who prefer traditional email signups over OAuth.

miurla:
## Summary
- Add Gravatar fallback for email/password users who don't have an OAuth-provided avatar
- Avatar URL is resolved server-side in `layout.tsx` using SHA-256 hashing and passed via props to `UserMenu`, avoiding client-side `useEffect`
- For `CurrentUserAvatar` (chat messages), Gravatar fallback is computed within the existing async hook
- Falls back gracefully: OAuth avatar → Gravatar → initials/placeholder icon (`?d=404`)

Closes #820

## Privacy note
Gravatar URLs contain a SHA-256 hash of the user's email address. While not directly reversible, the hash can be matched against known email lists to identify the original address. This is a common trade-off shared by services that use Gravatar (e.g. Stack Overflow, WordPress). The hash is exposed in the browser's network requests and rendered HTML for email/password users only — OAuth users are unaffected as their provider avatar takes precedence.

## Changes
| File | Change |
|---|---|
| `lib/utils/gravatar.server.ts` | Server-side Gravatar URL generator (Node.js crypto, sync) |
| `lib/utils/gravatar.ts` | Client-side Gravatar URL generator (Web Crypto API, async) |
| `app/layout.tsx` | Compute `avatarUrl` server-side and pass through props |
| `components/header.tsx` | Pass `avatarUrl` prop to `UserMenu` |
| `components/user-menu.tsx` | Use `avatarUrl` prop directly (no useEffect) |
| `hooks/use-current-user-image.ts` | Add Gravatar fallback in existing async effect |
| `lib/utils/__tests__/gravatar.test.ts` | Unit tests for Gravatar URL generation |

## Test plan
- [x] Unit tests pass (4 tests for hash generation, normalization, uniqueness)
- [x] Verified Gravatar URL returns HTTP 200 + image/png for registered email
- [x] Verified UI displays Gravatar avatar when OAuth avatar is absent
- [x] Lint, typecheck, format, and all 171 tests pass
- [ ] Verify email/password signup user sees Gravatar (if registered) or fallback initials
- [ ] Verify OAuth user still sees their OAuth avatar (not Gravatar)

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -7,6 +7,7 @@ import { getCurrentUserId } from '@/lib/auth/get-current-user'
 import { UserProvider } from '@/lib/contexts/user-context'
 import { createClient } from '@/lib/supabase/server'
 import { cn } from '@/lib/utils'
+import { getGravatarUrl } from '@/lib/utils/gravatar.server'
 
 import { SidebarProvider } from '@/components/ui/sidebar'
 import { Toaster } from '@/components/ui/sonner'
@@ -57,6 +58,7 @@ export default async function RootLayout({
   children: React.ReactNode
 }>) {
   let user = null
+  let avatarUrl: string | undefined
   const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL
   const supabaseAnonKey = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY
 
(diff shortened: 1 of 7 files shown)


miurla:
Closing this PR after further analysis of the privacy trade-off.

While Gravatar integration works correctly, exposing email hashes for **all** email/password users to benefit only the small subset who have Gravatar accounts is not a favorable trade-off. The existing initials-based fallback provides sufficient UX without any privacy cost.

See #820 for further discussion.

PR #846, opened 2026-05-22
fix(registry): route openai-compatible provider through chat completions endpoint

Tespera:
## Problem
With `OPENAI_COMPATIBLE_API_KEY` + `OPENAI_COMPATIBLE_API_BASE_URL=https://api.deepseek.com` (or any other OpenAI-compatible host — Moonshot, Zhipu, Together, etc.), every chat request returns 404:

```
url: 'https://api.deepseek.com/responses'
statusCode: 404
```

## Root cause
`lib/utils/registry.ts:14` builds the `openai-compatible` provider with `createOpenAI(...)`. In current `@ai-sdk/openai` versions that defaults to OpenAI's new Responses API (`/responses`). DeepSeek and the other major OpenAI-compatible hosts only implement the legacy `/v1/chat/completions` shape, so every dispatch 404s before reaching the model.

## Fix
Switch to `@ai-sdk/openai-compatible`'s `createOpenAICompatible`, which speaks `/v1/chat/completions` exclusively and accepts any host. Normalize the base URL so both `https://api.deepseek.com` and `https://api.deepseek.com/v1` resolve to the same endpoint.

## Files changed
- `lib/utils/registry.ts` — replace `createOpenAI(...)` with `createOpenAICompatible(...)`, add `normalizeOpenAICompatibleBaseURL`
- `package.json` / `bun.lock` — add `@ai-sdk/openai-compatible`

## Testing
- `bun lint` / `bun typecheck` pass
- Manual: DeepSeek chat replies correctly in dev (previously 404'd on every request)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -9,6 +9,7 @@
         "@ai-sdk/gateway": "^3.0.114",
         "@ai-sdk/google": "^3.0.73",
         "@ai-sdk/openai": "^3.0.63",
+        "@ai-sdk/openai-compatible": "^2.0.47",
         "@ai-sdk/react": "^3.0.182",
         "@aws-sdk/client-s3": "^3.850.0",
         "@json-render/core": "^0.16.0",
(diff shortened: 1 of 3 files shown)


miurla (approved):
Thanks for the fix. Looks good to me. CI is green, and switching the OpenAI-compatible provider to the chat completions endpoint is the right fix for providers like DeepSeek that do not support OpenAI's Responses API.

One small cleanup around keeping the SDK provider name separate from the UI display name can be handled as a non-blocking follow-up after merge.

PR #890, opened 2026-06-13
feat: add fastCRW search provider

us:
## What
Adds **fastCRW** as a search provider, alongside Tavily / SearXNG / Exa / Firecrawl. Selectable via `SEARCH_API=crw`.

## Why
[fastCRW](https://fastcrw.com) is a Firecrawl-API-compatible web engine in a **single ~8MB binary** — self-host free or managed cloud. Flat pricing (1 credit = 1 page; no 4x stealth surcharge, no billed-on-failure), free anti-bot stealth. For anyone on the Firecrawl provider it's a one-setting swap.

## Changes (additive only)
- `lib/crw/`: client + types (mirrors the sibling provider client).
- `lib/tools/search/providers/crw.ts` + registered in `providers/index.ts`.
- `.env.local.example`: `crw` added to `SEARCH_API`.
- `lib/tools/search/providers/__tests__/crw.test.ts`.

## Config
`SEARCH_API=crw`, `CRW_API_KEY` from https://fastcrw.com/dashboard (free tier); base URL overridable for self-host.

Happy to adjust — I maintain the integration and can provide free credits.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -74,6 +74,12 @@ TAVILY_API_KEY=[YOUR_TAVILY_API_KEY]
 # Get your API key at: https://www.firecrawl.dev/app
 # FIRECRAWL_API_KEY=[YOUR_FIRECRAWL_API_KEY]
 
+# Option 4: fastCRW (Firecrawl-compatible web scraper; single binary, self-host or cloud)
+# Get your API key at: https://fastcrw.com
+# CRW_API_KEY=[YOUR_CRW_API_KEY]
+# Override the base URL to point at a self-hosted instance (default: https://fastcrw.com/api)
+# CRW_API_URL=http://localhost:3000
+
 # Brave Search (Optional - video/image support for general searches)
 # Get your API key at: https://brave.com/search/api/
 # BRAVE_SEARCH_API_KEY=[YOUR_BRAVE_SEARCH_API_KEY]
@@ -113,7 +119,7 @@ ANONYMOUS_USER_ID=anonymous-user
 # Search Configuration
 # -----------------------------------------------------------------------------
 # Select search provider (default: tavily)
-# SEARCH_API=tavily  # Options: tavily, searxng, exa, firecrawl
+# SEARCH_API=tavily  # Options: tavily, searxng, exa, firecrawl, crw
 
 # SearXNG (Self-hosted)
 # SEARXNG_API_URL=http://localhost:8080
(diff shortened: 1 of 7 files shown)


miurla:
Thanks for the PR. The implementation is clean.

We're intentionally keeping the set of built-in search providers small and aren't adding new provider implementations right now. This is the same call we've made on recent provider contributions, so it's about maintenance surface, not fastCRW specifically.

Since fastCRW is Firecrawl-API-compatible, a separate `crw` provider isn't the right shape anyway. The equivalent would be a base-URL override on our existing `FirecrawlClient` (it currently hardcodes the endpoint). We don't have demand for that today, so it's not planned, but that's the direction if it comes up.

Closing for those reasons. Appreciate the contribution.

us:
Totally fair call on keeping the provider surface small — thanks for the thoughtful reply, and no worries on closing.

One clarification, because it changes both the maintenance math and the "just another Firecrawl-compatible thing" read: fastCRW isn't a hosted API or a client — it's a fully open-source engine (single ~8MB Rust binary, AGPL).

The part most people miss: **self-hosting Firecrawl's OSS barely gives you a working scraper.** Its real anti-bot/stealth path (fire-engine) is gated behind a cloud-only flag, so the self-hosted build falls back to plain fetch/Playwright and can't get through Cloudflare or most protected / JS-heavy sites — and you still have to run a multi-service stack (Redis + workers + Playwright) just to get that degraded version. fastCRW ships the whole thing in the open core: Cloudflare JS-challenge handling, UA rotation, SPA rendering, BYO-proxy + rotation, and an HTTP→headless→proxy fallback ladder — one binary, no cloud, no asterisks.

So "self-hosted Morphic + self-hosted fastCRW" is an actually-complete, cloud-free search+scrape stack — which is the one combination you basically can't build on Firecrawl's OSS today. That's the difference between deploying their open source (gets you very little) and deploying ours (gets you the real thing).

That said, you nailed the right shape regardless: instead of a new provider, just make the existing `FirecrawlClient` endpoint configurable — a `FIRECRAWL_API_BASE_URL` env defaulting to the current hardcoded URL. ~1 line, zero new surface, and it also helps anyone pointing at a self-hosted Firecrawl.

Would you be open to **reconsidering with that smaller shape**? If so I'll happily reopen/replace this with the one-line base-URL change — no new provider, nothing extra to maintain. Either way, thanks for taking the time.

PR #902, opened 2026-06-26
perf(rls): wrap current_setting() in (select …) for per-statement evaluation

dmitrymaranik:
## What

Wraps the `current_setting('app.current_user_id', true)` call in the four per-user RLS policies (`users_manage_own_chats`, `users_manage_own_notes`, `users_manage_own_files`, and the `USING` of `users_anonymize_own_feedback`) in a scalar subquery — `(select current_setting(…))` — in `lib/db/schema.ts`.

## Why

Postgres re-evaluates a bare function call in a policy's `USING` / `WITH CHECK` **once per row**. Wrapping it in `(select …)` makes it an InitPlan the planner evaluates **once per statement** and caches — identical result, far fewer calls on a large scan. It's the same optimization [Supabase documents](https://supabase.com/docs/guides/database/postgres/row-level-security) for `auth.uid()` / `current_setting()` in RLS.

```sql
-- before — evaluated per row:
USING (user_id = current_setting('app.current_user_id', true))
-- after — evaluated once per statement:
USING (user_id = (select current_setting('app.current_user_id', true)))
```

## Safety

Predicate-equivalent — `(select current_setting(…))` returns the same value, so **row visibility is unchanged**. I confirmed it with a static RLS analyzer ([pgrls](https://github.com/pgrls/pgrls)): its `PERF001` finding on these four policies clears to zero after the wrap, with nothing else changed.

The `EXISTS`-subquery policies (`users_manage_chat_messages`, `users_manage_message_parts`) already evaluate `current_setting` inside a subquery, so they're left untouched.

## Migration

This changes the policy source in `schema.ts`; the migration is best regenerated via your Drizzle flow (`drizzle-kit generate`) so the snapshot stays in sync. For reference, the equivalent SQL is `ALTER POLICY … USING ((select current_setting(…))) WITH CHECK (…)` on the four policies. Happy to add the generated migration to this PR if you'd prefer it here.

### DIFF
--- a/drizzle/0016_wrap_rls_current_setting.sql
+++ b/drizzle/0016_wrap_rls_current_setting.sql
@@ -1,0 +1,13 @@
+-- Custom SQL migration file, put your code below! --
+
+-- Wrap current_setting('app.current_user_id', true) in (select ...) in the per-user RLS
+-- policies so Postgres evaluates it once per statement (an InitPlan) instead of once per
+-- row. Predicate-equivalent (row visibility unchanged) -- it only changes when the value
+-- is computed. The EXISTS-subquery policies (messages, parts) are intentionally left as-is.
+-- Hand-written ALTER POLICY (not drizzle-generated) because the notes/files tables come
+-- from raw-SQL migrations 0012/0013 and aren't in the drizzle snapshot, so `drizzle-kit
+-- generate` would mis-emit CREATE TABLE/POLICY for them.
+ALTER POLICY "users_manage_own_chats" ON "chats" USING (user_id = (select current_setting('app.current_user_id', true))) WITH CHECK (user_id = (select current_setting('app.current_user_id', true)));--> statement-breakpoint
+ALTER POLICY "users_manage_own_files" ON "files" USING (user_id = (select current_setting('app.current_user_id', true))) WITH CHECK (user_id = (select current_setting('app.current_user_id', true)));--> statement-breakpoint
+ALTER POLICY "users_manage_own_notes" ON "notes" USING (user_id = (select current_setting('app.current_user_id', true))) WITH CHECK (user_id = (select current_setting('app.current_user_id', true)));--> statement-breakpoint
+ALTER POLICY "users_anonymize_own_feedback" ON "feedback" USING (user_id = (select current_setting('app.current_user_id', true)));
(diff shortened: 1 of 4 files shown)


miurla:
Thanks, the optimization is sound: predicate-equivalent, evaluated once per statement instead of per row, and leaving the `EXISTS`-subquery policies untouched is correct.

Please include the generated migration in this PR (the offer in your description), so `schema.ts` and the applied policies stay in sync. Then good to merge.

dmitrymaranik:
Done — pushed in the latest update (and rebased onto current `main`).

Tested on Postgres 16: the four `ALTER POLICY` statements apply cleanly and wrap all the per-user policies (`chats`, `files`, `notes`, `feedback` — `feedback`'s `WITH CHECK (user_id IS NULL)` is left as-is since it has no `current_setting`). Row visibility is unchanged, and the `EXISTS`-subquery policies stay untouched.

One note on the form: I added it as a **custom** raw-SQL migration (`0016`) rather than plain `drizzle-kit generate` output. `generate` doesn't give a clean diff here — the `notes`/`files` tables come from the raw-SQL migrations `0012`/`0013`, which aren't captured in the drizzle snapshot, so `generate` (even on a clean `main`) wants to emit `CREATE TABLE`/`CREATE POLICY` for them (a pre-existing snapshot drift, unrelated to this PR). To keep this PR scoped to just the policy wrap, I used `drizzle-kit generate --custom` and hand-wrote the four `ALTER POLICY`s. Happy to adjust the form if you'd prefer.

miurla (approved):
Migration looks correct. The four `ALTER POLICY` statements match the policies, `feedback`'s `WITH CHECK (user_id IS NULL)` is correctly preserved by altering `USING` only, and the `EXISTS`-subquery policies stay untouched. The `--custom` approach is the right call given the 0012/0013 snapshot drift. LGTM.

dmitrymaranik:
Thanks for merging — and for the careful review along the way. This came out of [pgrls](https://github.com/pgrls/pgrls), an open-source Postgres RLS linter/verifier; if it looks useful for keeping morphic's RLS policies tight, a star would help other folks find it. Happy to help wire it into CI if you'd ever want regression coverage on the policies.

PR #910, opened 2026-07-01
fix: handle bare string for include_domains/exclude_domains in Zod schema search tools

simply24365:
## Summary

AI models sometimes return `include_domains` / `exclude_domains` as a bare
string (`"reddit.com"`) instead of `string[]` (`["reddit.com"]`), causing
`TypeError: f.join is not a function` in `search-section.tsx`.

This was observed with the Xiaomi MIMO v2.5 model (OpenAI-compatible
endpoint), which does not enforce `strict: true` on function definitions.
The model can therefore produce type-invalid tool call arguments, which
Zod rejects at validation time — but the AI SDK still renders the raw
(unvalidated) input in the UI.

## Changes

1. **`lib/schema/search.tsx`** — Replaced `z.array(z.string())` with
   `z.union([z.string(), z.array(z.string())])` + `.transform()` so that
   bare strings are coerced into single-element arrays at Zod validation
   time, preventing the tool call from being rejected. Applied to both
   `searchSchema` and `strictSearchSchema` for `include_domains` and
   `exclude_domains`.

2. **`components/search-section.tsx`** — Added `Array.isArray` guard
   before `.join()` as defense-in-depth.

## Suggestion

This type of mismatch can affect any array-typed tool parameter. Consider
enabling `strict: true` on tools for providers that support it, or
implementing `repairToolCall` in `ToolLoopAgent` to automatically fix and
retry failed tool calls.

### DIFF
--- a/components/search-section.tsx
+++ b/components/search-section.tsx
@@ -63,7 +63,7 @@ export function SearchSection({
   const query = tool.input?.query || output?.query || ''
   const includeDomains = tool.input?.include_domains
   const includeDomainsString =
-    includeDomains && includeDomains.length > 0
+    Array.isArray(includeDomains) && includeDomains.length > 0
       ? ` [${includeDomains.join(', ')}]`
       : ''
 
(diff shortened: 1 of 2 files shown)


miurla (approved):
Thanks for the fix and the clear write-up. LGTM!

Issue #545, opened 2025-05-18
[BUG] Does not immediately redirect to new permalink


robertDouglass:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Vercel Runtime Logs

- [ ] I have checked the Vercel Runtime Logs for errors (if applicable)

### Current Behavior

I believe that the desired behaviour would be to immediately redirect to the new permalink of a new chat. It seems to wait to create the new chat entity until something else happens, eg. the streaming response is starting (or done?) 

I propose that it should immediately create and redirect to the new chat before sending the request off to LLM

### Expected Behavior

Submitting chat form creates new chat entity and updates the URL to that new entity before starting LLM queries

### Steps To Reproduce

See video https://youtu.be/iw_vV87mM-k

### Environment

```markdown
- OS:
- Browser:
```

PR #912, opened 2026-07-12
Fix a small issue in morphic (#545)

bglglzd:
Small scoped patch based on the reported behavior.

Related to #545.


miurla:
Thanks, but this change only adds a trailing newline to a comment-only file and is unrelated to #545 (which is about redirecting to the new chat permalink). Closing as it doesn't address the referenced issue.

Issue #927, opened 2026-08-02
Streaming request rejected with "Invalid body: failed to parse JSON value" on multi-turn chats


miurla:
## Problem

Some chat turns fail before the model produces anything: the provider rejects the streaming request with

```
Invalid body: failed to parse JSON value. Please check the value to ensure it is valid JSON.
(Common errors include trailing commas, missing closing brackets, missing quotation marks, etc.)
```

The error is recorded on `research-agent:ai.streamText` and its child `research-agent:ai.streamText.doStream`, so it fails at the first generation step and the user gets no answer. It is not caught and retried anywhere, so the turn is simply lost.

## Evidence

From production execution traces (a small number of occurrences within one chat, seconds to a minute apart, i.e. an initial failure plus a user retry):

- Model: `gpt-5.6-luna`, quick search mode. The error signature does not appear on the previously configured models in earlier traces, so it is new since the model switch in #923.
- Both failed on the very first generation of the turn, with zero successful steps.
- Both were multi-turn rather than first-turn requests. The serialized request input was comfortably inside the model's context window in both cases, so this is not a length limit.
- The captured request input contains no control characters and no U+FFFD replacement characters, so a simple invalid-UTF-8 payload is not obviously the cause.
- One structural oddity in both: every assistant history entry carries two `text` content parts rather than one.
- The larger of the two also had two consecutive `user` messages at the end.

Root cause is not identified. The provider only reports that the body failed to parse, and the raw outgoing body is not currently logged anywhere, so there is nothing to inspect.

## Proposed direction

1. On a provider 400, log the offending request body (the AI SDK surfaces it as `requestBodyValues` on `APICallError`) behind the existing error path, redacted as needed, so the next occurrence is diagnosable.
2. Check how assistant history entries are serialized for this provider, specifically why they carry two `text` parts, and whether consecutive same-role messages are valid for it.
3. If a malformed part can be identified, sanitize or normalize the history before the request rather than letting the turn fail.

## Acceptance criteria

- A provider 400 on the chat stream produces a log entry containing enough of the rejected body to identify the malformed portion.
- The cause of the two `text` parts per assistant history entry is either confirmed intentional or fixed.
- The failure is either resolved or reclassified with a concrete explanation.

miurla:
[...]

PR #945, opened 2026-08-06
fix(streaming): record provider API call diagnostics on the failing trace

miurla:
Refs #927

## Problem

Chat turns occasionally fail with a provider 400 whose only description is that the request body could not be parsed. #928 added `logAPICallErrorDiagnostics`, which builds a structural summary of the rejected request so the next occurrence would be diagnosable.

That summary is written with `console.error`, which lands in the platform's runtime logs. Their retention is minutes in practice, measured at roughly a 25 minute window. These failures are found by reading traces, typically hours after the fact, so by the time anyone looks the diagnostics are gone. A recurrence was investigated about 14 hours after it happened and nothing remained.

Meanwhile the failure itself is recorded in Langfuse and retained for days, but the root span carries only a generic `statusMessage` such as `bad_request: The request could not be processed.`, which is the same for every failure class and identifies nothing.

## Root cause

Not the 400 itself, which is still unexplained and stays open in #927. The defect fixed here is that the diagnostics built for it are written to a sink that cannot outlive the investigation window, so every occurrence is unusable by the time it is seen.

## Fix

- Extract the diagnostics construction from `logAPICallErrorDiagnostics` into an exported `buildAPICallErrorDiagnostics`, which returns the summary object or `null` when the error is not an `APICallError` or the summary cannot be built.
- In both stream paths, when the root span is already being marked `ERROR` for a stream failure, attach that summary as span metadata. `level` and `statusMessage` are unchanged.
- `logAPICallErrorDiagnostics` keeps its exact previous output, including the 4000 character truncation, and its existing call sites are untouched.

What gets recorded is unchanged from what #928 already computed: key names, serialized sizes, message roles, content part types and lengths, plus enumerated provider error fields. No request or response content. This adds no new exposure, since full model input and output already go to the same destination for every trace.

## Verification

- `bun run test`, `bun lint`, `bun typecheck`, `bun format:check`, `bun run build` all pass.
- New tests cover: `buildAPICallErrorDiagnostics` returning `null` for a plain `Error`, `null`, `undefined` and a string; the full structural summary for a realistic `APICallError`, asserted against concrete expected values rather than by recomputing the production logic; and a privacy guard that places a sentinel string inside a message part and asserts it does not appear anywhere in the serialized diagnostics.
- Existing coverage that `logAPICallErrorDiagnostics` still emits the identical console output is kept and tightened to an exact-string assertion.

The effect is not visible in any product metric. The concrete expectation is that the next occurrence of a provider `APICallError` on the chat stream produces a root `research` span whose metadata carries the rejected request's structure, making it diagnosable from traces alone.

### DIFF
--- a/lib/streaming/create-chat-stream-response.ts
+++ b/lib/streaming/create-chat-stream-response.ts
@@ -30,7 +30,10 @@ import {
   EMPTY_RESPONSE_STATUS_MESSAGE,
   isEmptyResponse
 } from './helpers/is-empty-response'
-import { logAPICallErrorDiagnostics } from './helpers/log-api-call-error'
+import {
+  buildAPICallErrorDiagnostics,
+  logAPICallErrorDiagnostics
+} from './helpers/log-api-call-error'
 import { persistStreamResults } from './helpers/persist-stream-results'
 import { prepareMessages } from './helpers/prepare-messages'
 import { stripSpecFromMessages } from './helpers/strip-spec-from-messages'
@@ -95,9 +98,11 @@ export async function createChatStreamResponse(
     const endTracing = async () => {
       if (rootSpan) {
         if (hasStreamError) {
+          const apiCallDiagnostics = buildAPICallErrorDiagnostics(streamError)
           rootSpan.update({
             level: 'ERROR',
-            statusMessage: describeStreamError(streamError)
+            statusMessage: describeStreamError(streamError),
+            ...(apiCallDiagnostics && { metadata: { apiCallDiagnostics } })
           })
         } else if (hasEmptyResponse) {
           rootSpan.update({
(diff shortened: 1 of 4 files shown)


miurla:
Note on checks: the GitHub Actions **CI** workflow did not start for this PR. No workflow run was created for the head commit by either the `opened` or a subsequent `reopened` event, and no run has been created anywhere in this repository for roughly the last 21 hours, with the most recent being on the previous day's merges. The workflow is `active` and repository Actions permissions are `enabled`, so this looks account or billing level rather than anything about this branch. Flagging it because it affects every PR right now, not just this one.

The Vercel checks passed. All five gates from the contributing requirements were run locally against this branch and passed: `bun run test` (357 passed, 3 skipped), `bun lint`, `bun typecheck`, `bun format:check`, and `bun run build`.

miurla:
Correction to the note above: the CI workflow did fire after all. It was delayed rather than blocked, and all five GitHub Actions checks are now green on this head commit (Lint, Type Check, Format Check, Test, Build), alongside the Vercel checks. Disregard the account or billing theory.

PR #947, opened 2026-08-07
fix: deduplicate repeated attachments instead of replaying every copy

### LINKED ISSUE #946: Deduplicate repeated attachments instead of replaying every copy
## Problem

The same file can end up attached more than once in a chat — most commonly when a user re-uploads because the first attachment was not acknowledged in the reply. Every copy is then replayed to the model on each subsequent turn.

Attachments are not cheap: a single PDF can expand to several hundred thousand tokens. So one accidental duplicate roughly doubles the input of every turn that follows it, for no added information.

## Proposed direction

When building the model messages, drop attachments that are byte-identical to one already present in the same conversation, keeping the first occurrence. The R2 object key is the natural identity (falling back to filename + size).

This is separate from `capHistoricalAttachments`, which bounds attachment **count**; duplicates can sit far below that cap and still dominate the payload.

## Acceptance criteria

- A conversation containing the same file attached twice sends it to the model once.
- Distinct files with the same name are not collapsed.

miurla:
Closes #946

The same file can end up attached more than once in a chat, most often when a user re-uploads because the first attachment went unmentioned in the reply. Every copy was then replayed to the model on each later turn, for no added information.

## Changes

**Send a repeated attachment once** (`lib/streaming/helpers/dedupe-attachments.ts`)

When building the model messages, an attachment identical to one already present collapses to a short placeholder and only the first occurrence is sent. Identity is the R2 object key, falling back to the url for parts stored without one; the filename is deliberately not part of it, so two different files both called `screenshot.png` stay separate.

Applied after `capHistoricalAttachments`, so an attachment that was already turned into a history placeholder cannot swallow the copy the user just re-attached. The decision for any part depends only on the parts before it, so appending turns never rewrites the prompt prefix and the provider cache survives.

**Point a repeated upload at the copy already stored** (`app/api/upload/route.ts`)

Uploading the same file again minted a new object key, which reads as a different file everywhere downstream. The upload now looks for a stored candidate in the same chat and reuses its key, so no second object and no second library row are written.

The lookup is scoped to one chat because object keys live under a per-chat prefix and a chat deletion wipes that prefix; reusing a key across chats would let one chat's deletion break another's attachment.

**Confirm the bytes before reusing** (`lib/storage/r2-client.ts`)

Name, media type and size only narrow the search to one candidate. The decision is a digest comparison, because handing back a different file under the same name would be a silent substitution, which is worse than the duplicate this avoids. S3 and R2 set the ETag of a single-part upload to the MD5 of the body, and everything here is written by one `PutObject` under a 5MB cap, so the `HeadObject` already being made carries the stored digest. A multipart or encrypted ETag is not a digest and is refused.

Every unconfirmed case falls through to a normal upload: no match, different bytes, missing object, unusable ETag, failed lookup, or anonymous mode.

## Notes

The two halves work together. Upload-side reuse alone saves storage but not tokens, since the same key would still be attached twice; message-side dedup alone rarely fires, because re-uploads used to produce distinct keys.

Separate from `capHistoricalAttachments`, which bounds attachment count: duplicates can sit far below that cap and still dominate the payload.

## Tests

- `lib/streaming/helpers/__tests__/dedupe-attachments.test.ts`: re-attach, same-message duplicates, same-name distinct files, url fallback, unidentifiable parts, attachment-only messages, prefix stability, placeholder escaping.
- `app/api/upload/__tests__/route.test.ts`: reuse, candidate narrowing, differing bytes, missing digest, failed lookup, anonymous mode.
- `lib/storage/__tests__/r2-client.test.ts`: ETag digest parsing and rejection.

lint, typecheck, format, build and the full suite pass.

### DIFF
--- a/app/api/upload/__tests__/route.test.ts
+++ b/app/api/upload/__tests__/route.test.ts
@@ -21,12 +21,14 @@ vi.mock('@/lib/storage/r2-client', () => ({
     Promise.resolve(`https://uploads.example.com/${key}?sig=abc`)
   ),
   isObjectStorageConfigured: vi.fn(() => true),
-  objectExists: vi.fn(() => Promise.resolve(true)),
+  getObjectContentMd5: vi.fn(() => Promise.resolve<string | null>(null)),
   R2_BUCKET_NAME: 'test-bucket'
 }))
 
+import { createHash } from 'node:crypto'
+
 import * as dbActions from '@/lib/db/actions'
-import { objectExists } from '@/lib/storage/r2-client'
+import { getObjectContentMd5 } from '@/lib/storage/r2-client'
 
 import { POST } from '../route'
 
@@ -64,11 +66,15 @@ function uploadRequest(bytes = 1234) {
   } as any
 }
 
+function md5OfUpload(bytes = 1234) {
+  return createHash('md5').update(Buffer.alloc(bytes)).digest('hex')
+}
+
 describe('POST /api/upload', () => {
   beforeEach(() => {
     vi.clearAllMocks()
     delete process.env.ENABLE_AUTH
-    vi.mocked(objectExists).mockResolvedValue(true)
+    vi.mocked(getObjectContentMd5).mockResolvedValue(md5OfUpload())
     vi.mocked(dbActions.createLibraryFile).mockResolvedValue({
       ...EXISTING,
       id: 'file-2',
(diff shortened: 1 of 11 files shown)


miurla:
Both review points addressed in d607af7.

**Include files uploaded before chat creation** — correct, and it hit the common case. The lookup now matches on the object key prefix instead of `chat_id`, since the key always carries the chat it was uploaded under while the column does not: a file attached to a brand-new chat is uploaded before the chat row exists and is stored with a null `chat_id`, and deleting a chat nulls it too. The per-chat confinement that motivated the original scoping is preserved, because the prefix is the thing a chat deletion wipes.

**Compare every matching metadata candidate** — also correct. The lookup returns several candidates newest-first and each is compared until one matches, so two same-sized versions of a file no longer push each other out of the search.

PR #959, opened 2026-08-12
fix(openai): set per-chat prompt cache key

### LINKED ISSUE #957: Prompt cache is not credited across turns: first call of each turn hits only the system prompt
## Summary

On long conversations, the **first model call of every turn reads back a prompt-cache hit of only about the system-prompt size**, while every subsequent call inside the same turn hits 94-99%. The cached-token count on those first calls is not merely low, it is *constant* — the same value repeats turn after turn, and it matches the length of the instructions block. Everything after the system prompt is re-billed at full input price on every turn.

This is not the prefix-instability class of bug fixed in #932/#933: the message array we send is byte-identical across turns for the entire shared region (verified by hashing each message). The prefix is stable and is simply not being credited.

## Evidence

Measured over one full production day on the `research-agent` generations (`gpt-5.6-luna`, OpenAI Responses API):

- Among generations with a context of 50k tokens or more, **~15% read a cache hit below 10%**, and those account for **~46% of the input spend of that whole group**.
- On the affected calls the cached-token value collapses onto a **small set of constants** (the two most common account for 19 of 22 occurrences), each corresponding to an instructions block length. It does not vary with the 160k+ tokens of history that follow.
- The same thread alternates cleanly: first call of a turn ~3% hit, the 2-3 calls that follow it in the same turn 94-99% hit, then the next turn drops back to ~3%.
- Two consecutive turns **25 seconds apart** both missed. TTL expiry and eviction do not explain a constant.

Structural check on two consecutive requests (end of turn N vs start of turn N+1):

```
idx   end of turn N                     start of turn N+1
...   (identical, hashed message by message)
 46   assistant[text]                   assistant[text]
 47   user[text]                        user[text]        <- first divergence
 48   assistant[text|tool-call]         -
 49   tool[tool-result]                 -
 ...
```

The shared prefix is ~161k tokens. The reported cache hit is ~5k.

## Suspected cause

We never send a prompt cache key. `prompt_cache_key` does not appear anywhere in the codebase, and `providerOptions` is populated only from the model config.

OpenAI's prompt caching routes a lookup using the prefix hash **plus** `prompt_cache_key`. With no key supplied, requests from the same conversation are free to land on different backends, and a backend that does not hold the entry can only serve the small shared head. That matches the observed shape precisely: a constant, instructions-sized hit rather than a partial hit that scales with the shared history.

`@ai-sdk/openai` (v3.0.63, already installed) exposes this as `promptCacheKey`:

```ts
// node_modules/@ai-sdk/openai/dist/index.d.ts
promptCacheKey?: string | undefined;
```

## Proposed fix

Pass a stable per-conversation `promptCacheKey` (the chat id is the natural value) into the agent's `providerOptions.openai`.

`createResearcher` ([lib/agents/researcher.ts](https://github.com/miurla/morphic/blob/main/lib/agents/researcher.ts)) currently receives `{ model, modelConfig, searchMode }` and has no access to the chat id, so this needs a small amount of plumbing from the stream-response builders. Merge the key into any `providerOptions` already coming from the model config rather than replacing them.

## Why it matters

The cost of a turn on a long thread is dominated by this. Because the miss re-bills the entire accumulated history, the penalty grows with conversation length, so it is worst exactly where a research product is most valuable: deep threads that carry pasted documents and accumulated search results. A thread that has ratcheted up to ~166k tokens of context pays that in full on every single turn, while the identical content is served at cache rates within the turn.

## How to verify

Compare, for the same thread shape, the cached-token count on the first generation of each turn before and after the change. Success is that value scaling with the shared history instead of pinning to the instructions length. Note the two existing read disciplines still apply: split file-bearing from file-free generations, and exclude the single heaviest session, since a blended cache rate moves with traffic mix on its own.

⚠️ One turn will still legitimately miss when history attachments are dropped by the #942 / #951 bounds, since that rewrites the prefix by design. Exclude turns carrying the omission placeholders when measuring.

sushanttiwari-14:
## Problem

The first OpenAI model call of each turn can miss most of the reusable prompt cache, causing long conversation histories to be billed again at the full input rate.

The researcher agent did not provide a stable `promptCacheKey`, even though the chat ID is available in both stream-response paths.

## Fix

- Pass the chat ID from persistent and ephemeral chat streams into the researcher.
- Set the chat ID as `providerOptions.openai.promptCacheKey`.
- Preserve existing OpenAI options and other provider options.
- Leave non-OpenAI providers unchanged.

## Verification

- Added unit coverage for merging the cache key with existing provider options.
- Verified non-OpenAI provider options remain unchanged.
- Added coverage confirming both stream paths pass the chat ID.
- `bun lint`
- `bun typecheck`
- `bun format:check`
- `bun run test` — 478 tests passed
- `bun run build`

The production cache-hit improvement should be confirmed after deployment using the first-generation-per-turn measurements described in the issue.

Fixes #957

### DIFF
--- a/lib/agents/__tests__/researcher.test.ts
+++ b/lib/agents/__tests__/researcher.test.ts
@@ -1,0 +1,117 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type { Model } from '@/lib/types/models'
+
+const mocks = vi.hoisted(() => ({
+  agentOptions: undefined as Record<string, any> | undefined
+}))
+
+vi.mock('ai', () => ({
+  stepCountIs: vi.fn(),
+  tool: vi.fn(options => options),
+  ToolLoopAgent: vi.fn(function (
+    this: { tools: unknown },
+    options: Record<string, any>
+  ) {
+    mocks.agentOptions = options
+    this.tools = options.tools
+  })
+}))
+
+vi.mock('@/lib/tools/fetch', () => ({
+  fetchTool: {}
+}))
+
+vi.mock('@/lib/tools/question', () => ({
+  createQuestionTool: vi.fn(() => ({}))
+}))
+
+vi.mock('@/lib/tools/search', () => ({
+  createSearchTool: vi.fn(() => ({}))
+}))
+
+vi.mock('@/lib/tools/todo', () => ({
+  createTodoTools: vi.fn(() => ({ todoWrite: {} }))
+}))
+
+vi.mock('@/lib/utils/registry', () => ({
+  getModel: vi.fn(() => ({}))
+}))
+
+vi.mock('@/lib/utils/telemetry', () => ({
+  isTracingEnabled: vi.fn(() => false)
+}))
+
+vi.mock('../prompts/search-mode-prompts', () => ({
+  getAdaptiveModePrompt: vi.fn(() => 'Adaptive prompt'),
+  QUICK_MODE_PROMPT: 'Quick prompt'
+}))
+
+import { createResearcher } from '../researcher'
+
+function createModel(
+  providerId: string,
+  providerOptions?: Model['providerOptions']
+): Model {
+  return {
+    id: 'model-id',
+    name: 'Model',
+    provider: 'Provider',
+    providerId,
+    providerOptions
+  }
+}
+
+describe('createResearcher', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    mocks.agentOptions = undefined
+    vi.spyOn(console, 'log').mockImplementation(() => {})
+  })
+
+  it('adds the chat ID to existing OpenAI provider options', () => {
+    createResearcher({
+      model: 'openai:model-id',
+      modelConfig: createModel('openai', {
+        openai: { reasoningEffort: 'low' },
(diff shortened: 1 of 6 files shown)


miurla:
Thanks for picking this up, and for turning it around so quickly.

LGTM. What I checked:

- **Both call sites are covered.** There are exactly two `researcher` callers, `create-chat-stream-response.ts` and `create-ephemeral-chat-stream-response.ts`, and the chat id was already in scope in each, so nothing is left unkeyed.
- **Existing provider options survive.** The merge spreads both the top level and the nested `openai` object rather than replacing them, and the tests cover that alongside the non-OpenAI and missing-chat-id branches.
- **The key reaches the provider.** `@ai-sdk/openai` 3.0.63 maps `providerOptions.openai.promptCacheKey` onto `prompt_cache_key` on the Responses request.

I also ran the full gate set locally, since the workflow run on this fork was still waiting on approval: `bun lint`, `bun typecheck`, `bun format:check`, `bun run test` (478 passed, 3 skipped) and `bun run build` all pass.

One note aimed at reading the result rather than at this PR: the effect can only be confirmed in production, and the signal to look at is the cached-token count on the first generation of each turn, not a blended cache rate. We track that separately.

PR #964, opened 2026-08-15
fix(search): surface complete Brave failures

### LINKED ISSUE #963: Brave provider swallows its own search failures, so they cannot be classified or surfaced
## Problem

`BraveSearchProvider.search()` never fails. Its three sub-searches (`searchWeb`, `searchVideos`,
`searchImages`) each wrap their body in a `try/catch` that only calls `console.error`, so any
failure is consumed there. `search()` then resolves normally with whatever slice of `results` was
populated, which for a total failure is an empty result set.

The effect is that a Brave outage, an auth failure, or a rate limit is indistinguishable from a
query that genuinely had no results. `lib/tools/search.ts` never sees an error, so
`ToolFailureError` is never constructed, nothing is classified, and the tool span records a
success. Tavily and SearXNG do not behave this way: they throw out of `search()` and are
classified normally.

## Evidence

In `lib/tools/search/providers/brave.ts`, each sub-search has the same shape:

```js
private async searchWeb(query, maxResults, results): Promise<void> {
  try {
    const response = await fetch(...)
    if (!response.ok) {
      console.error(`Brave web search failed: ${response.statusText}`)
      throw ...            // caught two lines below
    }
    ...
  } catch (error) {
    console.error('Brave web search error:', error)
  }
}
```

`search()` runs the selected sub-searches with `Promise.all` and returns `results` unconditionally.
Since every sub-search resolves, `Promise.all` always resolves.

This was surfaced by the Codex review on #961, which made the thrown errors carry their HTTP status
so the existing classifier can read them. That fix works end to end for Tavily and SearXNG; for
Brave the status is now present on the error but still never escapes the provider. The swallowing
predates that PR and is unchanged by it.

Configure Brave as the provider, make its endpoint return a 4xx or 5xx, and issue a search. The
tool call completes successfully with no results rather than raising a tool failure, and nothing
in the trace distinguishes it from a genuine empty result. Read `lib/tools/search/providers/brave.ts`
alongside `lib/tools/search/providers/tavily.ts` for the contrast.

## Proposed direction

This needs a decision about partial-failure semantics before any code changes, which is why it is
filed rather than fixed:

1. When more than one content type is requested and only some fail, decide whether the search
   should fail as a whole or return the partial results it did get. Returning partial results is
   probably right for a mixed web/video/image request, but that is a product call.
2. When **every** requested sub-search fails, the provider should surface a failure rather than an
   empty success. This is the case with no real argument against it, and it is what makes an outage
   visible.
3. Whichever is chosen, the error that propagates should be the one carrying the HTTP status from
   #961, so the existing `classifyToolFailure` path produces the right copy and the right
   `retryable` flag without further changes.

Worth noting for prioritisation: Brave is an optional provider in this project, not the default, so
the blast radius is limited to deployments that configure it.

## Acceptance criteria

- A Brave search in which every requested content type fails surfaces a tool failure instead of an
  empty successful result.
- That failure is classified by HTTP status, matching how Tavily and SearXNG failures are already
  classified.
- The chosen behaviour for a partial failure is deliberate and covered by a test, rather than a
  side effect of where the `try/catch` happens to sit.

Refs #961.

sushanttiwari-14:
## Problem

The Brave search provider catches and logs failures from its web, video, and image requests without propagating them. When every requested content type fails, the provider therefore returns an empty successful result instead of a tool failure.

This makes provider outages, authentication failures, and rate limits indistinguishable from a valid search with no results. It also prevents the existing HTTP-status classifier from producing the appropriate user-facing error and retry behavior.

## Fix

- Allow each Brave sub-search to rethrow its status-bearing error after logging it.
- Wait for all requested sub-searches with `Promise.allSettled`.
- Return partial results when at least one requested content type succeeds.
- Propagate the first failure when every requested content type fails, allowing the existing tool-error path to classify it by HTTP status.

## Verification

- Added coverage confirming a complete Brave failure is surfaced and classified by HTTP status.
- Added coverage confirming successful partial results are preserved when another requested content type fails.
- `bun lint`
- `bun typecheck`
- `bun format:check`
- `bun run test` — 489 tests passed, 3 skipped
- `bun run build`

Fixes #963

### DIFF
--- a/lib/tools/search/providers/__tests__/brave.test.ts
+++ b/lib/tools/search/providers/__tests__/brave.test.ts
@@ -1,0 +1,99 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import {
+  getToolFailureMessage,
+  serializeToolFailure,
+  ToolFailureError
+} from '@/lib/errors/tool-error'
+
+import { BraveSearchProvider } from '../brave'
+
+describe('BraveSearchProvider', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs()
+    vi.unstubAllGlobals()
+    vi.restoreAllMocks()
+  })
+
+  it('surfaces a classified failure when every requested search fails', async () => {
+    vi.stubEnv('BRAVE_SEARCH_API_KEY', 'test-key')
+    vi.stubGlobal(
+      'fetch',
+      vi.fn(async () =>
+        Promise.resolve(
+          new Response(null, {
+            status: 429,
+            statusText: 'Too Many Requests'
+          })
+        )
+      )
+    )
+    vi.spyOn(console, 'error').mockImplementation(() => undefined)
+
+    const error = await new BraveSearchProvider()
+      .search('query', 5, 'basic', [], [], {
+        content_types: ['web', 'image']
+      })
+      .catch(error => error)
+
+    expect(error).toBeInstanceOf(Error)
+    expect(error.status).toBe(429)
+
+    const toolError = new ToolFailureError('search', error)
+    expect(getToolFailureMessage(toolError)).toBe(
+      'The search service is rate limiting requests. Please try again shortly.'
+    )
+    expect(JSON.parse(serializeToolFailure(toolError))).toMatchObject({
+      error:
+        'The search service is rate limiting requests. Please try again shortly.',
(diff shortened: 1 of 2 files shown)


miurla:
Thanks! LGTM.

PR #966, opened 2026-08-16
fix: stabilize related questions in deep chats

### LINKED ISSUE #944: Spec block emission degrades with conversation depth; replayed history is never bounded in practice
## Problem

The related-questions spec block is specified as expected on every substantive answer, but the rate at which the model actually emits it falls as conversation depth grows. Early turns follow the instruction most of the time; deep turns mostly do not.

The same mechanism has a second cost: every turn replays the entire prior history, including all tool output, so a long thread's per-request input keeps growing with no effective ceiling.

## Evidence

Measured on persisted assistant messages in adaptive search mode, grouping turns by their ordinal position within their chat and testing whether the answer contains a related-questions `spec` fence.

The decline is **monotonic in depth**: the first turn in a chat follows the instruction most of the time, and compliance falls at every subsequent depth band, dropping to a small fraction of the first-turn rate by the time a thread reaches its seventh turn and beyond. The shape reproduces across multiple consecutive days, including days with no prompt change in flight, so it is not an artifact of any single prompt regression.

A corollary worth recording for anyone measuring this: because compliance depends this strongly on depth, a blended emission rate also moves whenever the mix of shallow and deep turns moves, independent of prompt quality. Depth-stratified cuts are the only reliable read.

History is not bounded in practice:

- `shouldTruncateMessages` (`lib/utils/context-window.ts:296`) only fires above `getMaxAllowedTokens`, which is derived from the model's configured `contextWindow` minus its `outputTokens` minus a 10% safety buffer. For a model configured with a very large context window, that threshold sits orders of magnitude above the input size real conversations actually reach, so the truncation path effectively never runs and the full history is replayed on every turn.
- `compactHistoricalMessages` (`lib/streaming/helpers/compact-historical-messages.ts`) bounds only per-turn source context, at `MAX_SOURCE_CONTEXT_CHARS = 800`.
- `capHistoricalAttachments` bounds only attachments.

None of the three bounds replayed tool output, which is the bulk of a deep thread.

### How to reproduce this measurement

Group persisted assistant messages by their position within their chat, then within each depth band compute the share whose text contains a related-questions `spec` fence. Compare bands within a single day so that model, prompt and traffic mix are held constant.

## Proposed direction

Several approaches compete and the trade-off is a product decision, so this is an issue rather than a PR:

1. Bound replayed history by a budget set well below the model's context window instead of by the window itself, so deep threads converge to a stable prompt size.
2. Prefer dropping or summarizing older tool output ahead of older conversational turns, since tool output dominates the payload and is the least reusable part of it.
3. Independently, make the spec instruction's position relative to the history stable, for example by restating it after the history rather than only in the system prefix.

Options 1 and 2 also address the unbounded input growth. Option 3 targets emission alone and is testable on its own.

## Acceptance criteria

- Emission at the deepest bands comes materially closer to the first-turn rate on comparable traffic. Target levels are tracked internally rather than stated here.
- A deep thread's per-turn input size stops growing without bound.
- Whichever option is taken, the effect is verified on a depth-stratified cut rather than on a blended rate, since a blended rate confounds prompt quality with turn-depth mix.

sushanttiwari-14:
## Problem

The related-questions instruction appears only near the beginning of the system prompt. As conversation history grows, the instruction becomes increasingly distant from the current request, causing related-question spec blocks to be emitted less reliably in deeper conversations.

## Fix

- Add a focused related-questions reminder next to the current user request.
- Apply the reminder through the shared researcher configuration so both persisted and ephemeral chats receive it.
- Preserve existing text and file parts without mutating the original messages.
- Keep history compaction, stored messages, tool output, and rendering unchanged.
- Add coverage for deep conversations, file attachments, missing user messages, and researcher integration.

## Verification

- `bun lint`
- `bun typecheck`
- `bun format:check`
- `bun run test` — 497 tests passed, 3 skipped
- `bun run build`

Fixes #944

### DIFF
--- a/lib/agents/__tests__/researcher.test.ts
+++ b/lib/agents/__tests__/researcher.test.ts
@@ -2,6 +2,8 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import type { Model } from '@/lib/types/models'
 
+import { RELATED_QUESTIONS_REMINDER } from '../prompts/related-questions-reminder'
+
 const mocks = vi.hoisted(() => ({
   agentOptions: undefined as Record<string, any> | undefined
 }))
@@ -112,6 +114,30 @@ describe('createResearcher', () => {
 
     expect(mocks.agentOptions?.providerOptions).toEqual({
       openai: { reasoningEffort: 'low' }
+    })
+  })
+
+  it('places the related-questions reminder after conversation history', () => {
+    createResearcher({
+      model: 'openai:model-id',
+      modelConfig: createModel('openai')
+    })
+
+    const prepareCall = mocks.agentOptions?.prepareCall
+    const result = prepareCall({
+      messages: [
+        { role: 'user', content: 'Initial question' },
+        { role: 'assistant', content: 'Initial answer' },
+        { role: 'user', content: 'Deep follow-up' }
+      ]
+    })
+
+    expect(result.messages.at(-1)).toEqual({
+      role: 'user',
+      content: [
+        { type: 'text', text: 'Deep follow-up' },
+        { type: 'text', text: RELATED_QUESTIONS_REMINDER }
+      ]
     })
   })
 })
--- a/lib/agents/prompts/__tests__/related-questions-reminder.test.ts
+++ b/lib/agents/prompts/__tests__/related-questions-reminder.test.ts
@@ -1,0 +1,68 @@
+import type { ModelMessage } from 'ai'
+import { describe, expect, it } from 'vitest'
+
+import {
+  appendRelatedQuestionsReminder,
+  RELATED_QUESTIONS_REMINDER
+} from '../related-questions-reminder'
+
+describe('appendRelatedQuestionsReminder', () => {
+  it('places the reminder after the current user text in a deep conversation', () => {
+    const messages: ModelMessage[] = [
+      { role: 'user', content: 'First question' },
+      { role: 'assistant', content: 'First answer' },
+      { role: 'user', content: 'Follow-up question' }
+    ]
+
+    const result = appendRelatedQuestionsReminder(messages)
+
+    expect(result).not.toBe(messages)
+    expect(result.at(-1)).toEqual({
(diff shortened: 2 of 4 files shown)


miurla:
Thanks :+1:

I re-ran the depth-stratified measurement before merging, and the effect this PR targets reproduces under three controls: restricting to clearly substantive answers, so the documented skip cases can't account for it; comparing only within chats that reach the deeper bands, so it isn't a chat or user mix effect; and holding the model constant, so it isn't a model mix effect. Quick mode degrades at least as much as adaptive, so applying this through the shared researcher configuration rather than to one mode is the right call.

On the implementation, `prepareCall` is the right hook. It runs once per `agent.stream()` call, so the reminder is added exactly once per model call and does not accumulate across the tool loop, and it operates on the converted `ModelMessage[]`, so nothing reaches the persisted history. Moving this to `prepareStep` would produce an identical prompt under the same "append to the last user message" rule, so there is no reason to change it.

Two follow-ups, neither blocking:

1. The reminder restates the question count and the skip-case carve-out in its own words, duplicating what `getRelatedQuestionsSpecPrompt()` in `lib/render/prompt.ts` already specifies. If that prompt changes, the reminder drifts silently. Worth sourcing the shared parts from one place, or adding a test that ties the two together.

2. A competing mechanism is worth ruling out separately. `stripSpecFromMessages` removes the spec block from every historical assistant message, so by the seventh turn the model's context holds several examples of its own answers ending without one. That is a negative few-shot pressure that grows monotonically with depth, and it fits the observed shape as well as instruction distance does. If the reminder alone does not close the gap, keeping a short placeholder in place of the stripped block is the next thing to try.

Merging on the strength of the change being self-contained and low-risk. Per the acceptance criteria in #944, the effect still needs to be confirmed on a depth-stratified cut once this has live traffic, so that issue stays open until that read is in.

sushanttiwari-14:
Thanks for the detailed review! I’ll follow #944 for the production results and any follow-up work.

PR #968, opened 2026-08-17
feat(analytics): record whether the feedback control was offered

miurla:
## Problem

Thumbs-up / thumbs-down is the only sentiment signal the app collects, and the control renders only inside `{traceId && ...}` in `components/message-actions.tsx`. The trace id reaches the component as `metadata?.traceId` (`components/answer-section.tsx`), attached server-side on the stream's `start` part (`lib/streaming/create-chat-stream-response.ts`, `lib/streaming/create-ephemeral-chat-stream-response.ts`).

So whenever the trace id does not reach the client, the whole feedback control disappears from the answer and nothing records that it happened. A change in how many scores arrive is therefore ambiguous between a change in what users do and the control simply not being rendered, and nothing we collect can tell those apart. Clicks are recorded; impressions are not, so there is no denominator to read clicks against.

## Root cause

There is no instrumentation on the render side of the feedback control, only on the submit side. Absence of scores and absence of the control produce identical data.

## Fix

Emit one `feedback_control_shown` client event per assistant message, once the actions row is actually presented (visible, and the message is no longer streaming), carrying `hasTraceId`, `chatId` and `isGuest`. A ref keyed on `messageId` keeps re-renders, hover, parent state changes and React strict-mode double effects from duplicating it.

The change is additive. Rendering, the feedback submission path, and every existing event are untouched. No message text and no trace id value are sent.

## Verification

- New tests in `components/__tests__/message-actions.test.tsx`: one impression when the row becomes visible, nothing while streaming, no duplicate under strict mode or on parent re-render, and `hasTraceId: false` recorded for a message presented without a trace id.
- `bun lint`, `bun typecheck`, `bun format:check`, `bun run test` all pass locally.

Once this is live, "the control was never offered" and "the control was offered and not used" become separable, which is what makes any later read of feedback volume interpretable.

### DIFF
--- a/components/__tests__/message-actions.test.tsx
+++ b/components/__tests__/message-actions.test.tsx
@@ -1,0 +1,110 @@
+import React from 'react'
+
+import { render } from '@testing-library/react'
+import { beforeEach, describe, expect, test, vi } from 'vitest'
+
+import { captureClient } from '@/lib/analytics/posthog-client'
+
+import { MessageActions } from '../message-actions'
+
+vi.mock('@/lib/analytics/posthog-client', () => ({
+  captureClient: vi.fn()
+}))
+
+vi.mock('../library/library-context', () => ({
+  useLibrary: () => ({
+    openLibrary: vi.fn(),
+    upsertCachedNote: vi.fn()
+  })
+}))
+
+describe('MessageActions', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  test('records one feedback control impression when the row becomes visible', () => {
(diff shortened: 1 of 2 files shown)


miurla:
Addressed in 8709519. The dedupe key moved from a component-local ref to module scope, so reopening a conversation no longer re-counts every past answer, and the event now carries `hasFeedback` so impressions on an already-scored control can be excluded from the denominator rather than silently inflating it. Added a remount test that unmounts and re-renders the same message id, plus assertions for the new property.

@codex review

PR #974, opened 2026-08-18
fix(streaming): stabilize attachment weight boundary

### LINKED ISSUE #970: capHistoricalAttachments: the weight-budget boundary moves every turn, rewriting replayed history on the threads it is meant to make cheaper
## Problem

In an attachment-heavy thread, the set of historical attachments replaced by `[Attachment omitted from history: ...]` grows by one on every turn. Because that boundary sits near the START of the conversation, each turn rewrites the prompt prefix and the provider's cache is dropped for the whole request. These are exactly the threads the file exists to make cheaper, and they are the most expensive ones per request.

Comparing the prompts of consecutive turns in one such thread, the first divergence walks forward one attachment at a time: each turn flips exactly one more historical attachment into the placeholder, and every turn after the first reads a cached prefix no larger than the instructions block.

## Root cause

`lib/streaming/helpers/cap-historical-attachments.ts` has two boundaries and only one of them is stable.

The count cap uses `getDroppedCount()`, which quantizes the dropped prefix to whole blocks of `limit`. Its comment states the reason plainly: an unquantized boundary would move by one on every turn and rewrite the prefix.

The weight pass does not quantize. It drops the oldest surviving attachment until `survivingTokens <= tokenBudget`, recomputed against the current history each turn. Once a thread is over budget, each new attachment raises `survivingTokens` and `weightDropCount` grows by roughly one per turn. The comment above that pass claims the boundary "only ever moves forward, and each crossing rewrites the prefix once instead of on every turn". The forward-only half is true; the once half is not, and that comment should be corrected along with the code.

## A fix that looks obvious and does not work

The natural move is to quantize the weight boundary the way the count cap does, rounding the weight-dropped count up to a whole block of `limit`. Simulated over a synthetic thread that appends one large PDF per turn, this does not stabilize the boundary at all.

The reason is the guard that at least one attachment must survive. `DEFAULT_ATTACHMENT_TOKEN_BUDGET` is 200,000 and `estimateAttachmentTokens` charges a non-image at `size / 3.9`, so a single large PDF can be a large fraction of the budget on its own. When only one or two attachments can fit, the survivor clamp binds on every turn, the clamp itself moves by one per turn as history grows, and the quantization is bypassed. The clamp binds precisely in the threads that matter.

The deeper point: with a hard token budget and single attachments that are a large fraction of it, no budget-respecting boundary can also be stable. Something has to give, and which one is a product call:

- **Accept budget overshoot for prefix stability.** Quantize on the budget axis with hysteresis: only move the boundary once the surplus exceeds a margin, then drop to below budget minus that margin. Stability is proportional to the margin, and the margin has to exceed a typical single attachment to buy anything, which means tolerating a context noticeably above the nominal budget between crossings.
- **Anchor the kept set to the start of the thread instead of the end.** The boundary then never moves as history grows, which is prefix-stable by construction. The current turn's attachments are already exempt (`i < historyEnd`), so the newest upload stays live either way. The cost is that mid-thread attachments get dropped while older ones are kept, which inverts what users would expect the model to still be holding.
- **Accept the churn** and treat these threads as uncacheable, on the grounds that the budget matters more than the cache. This is the current behaviour, but right now it is accidental rather than chosen, and the comment in the file asserts the opposite.

Worth factoring in when picking: `estimateAttachmentTokens` over-states real cost for PDFs, since PDF bytes are compressed binary, so the budget effectively fires earlier than its name suggests and the boundary churns sooner than the constant implies.

## Acceptance criteria

- In a thread that keeps adding attachments while over the token budget, the set of placeholdered attachments is identical across consecutive turns for a bounded stretch, rather than changing every turn. The test should assert the replayed prefix is stable, not just that a count matches.
- Whatever bound is chosen is stated in the code with its reason, so the next reader does not restore an unquantized boundary.
- At least one historical attachment still survives, and the current turn's attachments are still never capped.
- Behaviour is unchanged when the count cap is disabled.

sushanttiwari-14:
Closes #970

## Problem

The weight based attachment cap recalculated its drop boundary on every turn. In attachment heavy conversations, each new attachment caused another historical attachment to be replaced with a placeholder.

Because this boundary appears near the beginning of the conversation, the changing prefix invalidated the provider cache on consecutive requests.

## Solution

This change groups historical attachments into cumulative token weight blocks anchored to the start of the surviving history. The boundary advances only when another complete block becomes available.

Each block requires at least two attachments, so a single oversized attachment cannot move the boundary on consecutive turns. The newest complete block and any incomplete tail are retained.

Attachments from the current user turn remain uncapped. At least one historical attachment is always preserved, and weight limiting continues to work when the count cap is disabled.

This approach can temporarily exceed the nominal token budget. The overshoot is intentional because a strict budget cannot also provide a stable prefix when individual attachments consume a large share of that budget.

## Testing

Added regression coverage to confirm that the replayed prefix remains identical across consecutive over budget turns.

The tests also cover heavy attachments below the count limit, disabled count caps, oversized individual files, and current turn attachments.

Validation completed successfully:

`bun run test` with 539 tests passed and 3 skipped

`bun typecheck`

`bun lint`

`bun format:check`

`bun run build`

### DIFF
--- a/lib/streaming/helpers/__tests__/cap-historical-attachments.test.ts
+++ b/lib/streaming/helpers/__tests__/cap-historical-attachments.test.ts
@@ -44,6 +44,26 @@ function thread(turns: number, perTurn: number): UIMessage[] {
   ]).flat()
 }
 
+function pdfThread(turns: number, size = 500_000): UIMessage[] {
+  return thread(turns, 0).map(message => {
+    if (message.role !== 'user') return message
+
+    return {
+      ...message,
+      parts: [
+        {
+          type: 'file',
+          mediaType: 'application/pdf',
+          filename: `${message.id}.pdf`,
+          url: `https://example.com/${message.id}.pdf`,
+          size
+        },
+        ...message.parts
+      ]
+    } as unknown as UIMessage
+  })
+}
+
 function filenamesReaching(messages: UIMessage[]): string[] {
   return messages.flatMap(message =>
     message.parts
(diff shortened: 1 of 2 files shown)


miurla:
LGTM, and thanks for taking this one on.

Anchoring the blocks to the start of the surviving history is the right call: it makes the boundary a function of the prefix, so appending an attachment cannot move it. Verified against the previous behaviour on synthetic threads, the boundary now holds still between crossings instead of advancing every turn, a placeholdered attachment never comes back, and the count-cap-disabled path is unchanged.

One residual property, recorded as a note on #970 rather than as a change request here: the block minimum is a count, so it does not cover a thread whose single turn adds a budget's worth of weight across several attachments. That is a property of the option you picked, not a defect, and the note explains the knob if it ever needs picking up.

PR #975, opened 2026-08-18
feat(upload): accept MP4 so video-capable models can read a clip

octo-patch:
Reason: Uploaded attachments are forwarded to the selected model, but the upload path accepts only JPEG, PNG and PDF, so a clip cannot reach a model that reads video.

## Problem

Attachments become `file` parts and are handed to whichever model the chat is using, with no per-format branching on the way out. The formats that may enter, though, are fixed at the upload boundary: JPEG, PNG and PDF. A model that accepts video therefore cannot be given a clip — the file is refused before it is ever stored, so the capability is unreachable no matter which model is selected.

MP4 was not deliberately excluded. It was simply absent from the allowlist, and the content check added in #952 rejected the ISO container as a consequence.

## Changes

- `lib/storage/file-signature.ts` detects MP4 from the ISO base media `ftyp` box: the box type at bytes 4..8, then the brand at bytes 8..12 checked against the MP4-compatible brands. The brand check is what keeps the shared container from over-matching — QuickTime (`qt  `) and M4A audio ride in the same box, and accepting the box alone would move their failure from the upload to the provider request. `video/mp4` is ordered ahead of PDF because PDF is the one format matched by scanning a window rather than a fixed offset.
- `video/mp4` is added to the declared-type gate (shared with the signature module) and to the three client allowlists that guard the attachment menu, the standalone picker and the drag-and-drop zone. The 400 message that enumerates the readable formats is updated to match.
- `lib/utils/attachment-tokens.ts` estimates a clip from a frame-sampled fixed cost, the same reasoning already applied to images, instead of falling through to the byte-length branch. Byte length is the wrong model for video and not merely imprecise: a clip at the upload ceiling estimates over a million tokens, which exceeds the whole 200k attachment budget from #951 and would evict every other attachment replayed to the model, and would trip context-window truncation on the turn that carries it. The estimate is deliberately above a single image's.

The existing signature test asserted that the production ISO container is rejected; it now asserts the container is detected as video, which keeps the original regression covered — a clip named `image.jpg` is still not sent to the model as a broken image. The route test's "unsupported content" fixture was that same container, so it is replaced with bytes that open no supported format.

## Scope

The 5MB upload ceiling is unchanged, so this admits short clips rather than arbitrary video. Raising that limit is a storage and cost decision, not part of making the format reachable, and it is left alone deliberately. Attachment previews already render a non-image as a labelled chip and a link, so a clip shows up as `MP4` without new UI.

## Verification

- `bun run test` — 542 passed, 3 skipped, 61 files. New coverage: MP4 detection, the production container now detected as video, a non-MP4 brand and a truncated brand both rejected, a clip stored under its detected type across the object write and the library row, and the fixed video token estimate.
- `bun lint`, `bun typecheck`, `bun run format:check`, `bun run build` all pass.
- Detection was also exercised against the `ftyp` boxes real encoders write (`isomiso2avc1mp41`, `mp42mp41isomavc1`, `dashiso6avc1mp41`), all detected as `video/mp4`, with a QuickTime and an M4A header rejected and JPEG/PNG detection unchanged.

### DIFF
--- a/app/api/upload/__tests__/route.test.ts
+++ b/app/api/upload/__tests__/route.test.ts
@@ -78,6 +78,13 @@ function pdfUploadBytes(bytes = 1234) {
   return contents
 }
 
+function mp4UploadBytes(bytes = 1234) {
+  const contents = new Uint8Array(bytes)
+  contents.set([0x00, 0x00, 0x00, 0x20])
+  contents.set(Buffer.from('ftypisom', 'ascii'), 4)
+  return contents
+}
+
 function md5OfUpload(bytes = 1234) {
   return createHash('md5').update(pdfUploadBytes(bytes)).digest('hex')
 }
@@ -133,9 +140,9 @@ describe('POST /api/upload', () => {
   })
 
   it('rejects allowed declared types with unsupported content', async () => {
-    const contents = new Uint8Array([
-      0x00, 0x00, 0x00, 0x20, 0x66, 0x74, 0x79, 0x70, 0x69, 0x73, 0x6f, 0x6d
-    ])
+    // Bytes that open no supported format: not a JPEG, PNG, MP4 container or
+    // PDF header.
+    const contents = new Uint8Array(32)
 
     const response = await POST(uploadRequest(contents.byteLength, contents))
 
@@ -146,6 +153,30 @@ describe('POST /api/upload', () => {
     expect(dbActions.findChatFileCandidates).not.toHaveBeenCalled()
     expect(send).not.toHaveBeenCalled()
     expect(dbActions.createLibraryFile).not.toHaveBeenCalled()
+  })
+
+  it('stores a clip under its detected video type', async () => {
+    vi.mocked(dbActions.findChatFileCandidates).mockResolvedValue([])
+    const contents = mp4UploadBytes()
+
+    const { file } = await (
+      await POST(
+        uploadRequest(contents.byteLength, contents, {
+          name: 'clip.mp4',
+          type: 'video/mp4'
+        })
+      )
+    ).json()
+
+    expect(file.mediaType).toBe('video/mp4')
+    expect(send).toHaveBeenCalledWith(
+      expect.objectContaining({
+        input: expect.objectContaining({ ContentType: 'video/mp4' })
+      })
+    )
+    expect(dbActions.createLibraryFile).toHaveBeenCalledWith(
+      expect.objectContaining({ mediaType: 'video/mp4' })
+    )
   })
 
   it('stores a mislabelled file under the type its bytes say it is', async () => {
--- a/app/api/upload/route.ts
+++ b/app/api/upload/route.ts
@@ -81,7 +81,7 @@ export async function POST(req: NextRequest) {
         {
           error: 'Unsupported file content',
           message:
-            'The file contents are not a JPEG, PNG, or PDF even though the file name suggests otherwise.'
+            'The file contents are not a JPEG, PNG, MP4, or PDF even though the file name suggests otherwise.'
         },
         { status: 400 }
       )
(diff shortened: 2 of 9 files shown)


miurla:
@codex review

miurla:
Reviewed this with model routing in mind, and the finding changes the framing of the PR, so I would rather check the premise with you than hand you a fix list.

## Only one provider in this repo can carry a video part today

Verified against the versions in this tree. Each of these throws during prompt conversion, before any network call:

```
openai(responses): AI_UnsupportedFunctionalityError :: 'file part media type video/mp4' functionality not supported.
anthropic:         AI_UnsupportedFunctionalityError :: 'media type: video/mp4' functionality not supported.
openai-compatible: AI_UnsupportedFunctionalityError :: 'file part media type video/mp4' functionality not supported.
google:            reaches the API (request is built)
```

Ollama is image-only upstream. The gateway provider does forward the part, but the AI Gateway models API still does not expose a video-input tag (vercel/ai#9417 is open), so capability cannot be read from metadata either.

Worth noting for the "which model reads video" question: the open models that do read video (Qwen3-VL, GLM-4.6V) are not reachable through this stack either. vLLM serves them with a `video_url` / `input_video` content type, which is an extension of the OpenAI schema, and `openai-compatible` throws before the request is built.

So the capability this PR makes reachable is reachable on `google:` only.

## What that means for the current shape

The allowlist is global and nothing gates MP4 on the receiving model:

- Cloud pins the model to `gpt-5.6-luna` (`config/models/cloud.json`) and hides the selector, so on cloud every video upload fails.
- Self-host defaults to `gpt-5.4-mini` (`lib/config/default-model.ts`), so it fails there too unless the user has explicitly selected Gemini.
- The failure is not confined to the turn that carries the clip. The user message is persisted before the model call (`lib/streaming/helpers/prepare-messages.ts:133`, and the optimistic save at `:102` for a new chat), attachments are replayed every turn, and the replay cap keeps the newest 10. So the throw repeats on every later turn in that chat.

That is the shape of the incident #952 closed. Today the same clip is refused at upload with a 400 that names the problem. After this change it is accepted, stored, and then breaks the conversation with a generic error.

## Question

Given that video is reachable only on Google today: do you still want to land this, or should it be closed?

**If yes**, the gate belongs on the receiving model's provider rather than on the upload boundary:

- Reaching video is a property of the provider, not of the deployment. `openai`, `anthropic`, `openai-compatible` and `ollama` cannot carry a video part at all, whatever an operator configures, so a per-deployment opt-in flag would mostly let someone switch on a state that still fails. Allow `video/mp4` only when the selected model's provider can read it, which today means `google` alone.
- Keep that decision in one place, something like `canReadMediaType(providerId, mediaType)` in `lib/config`, so the day an `openai-compatible` video path exists (a `video_url` conversion for a vLLM-served Qwen3-VL) it is one edit instead of four allowlists.
- The composer can already see the selection: `selectedModel` is a readable cookie (`lib/config/model-selection-cookie.ts`), and `chat-panel.tsx` already reads a cookie this way for search mode. Drop `video/mp4` from `accept` and from the drop-zone validation when the selected model cannot read it, and leave it out entirely when Google is not an enabled provider (`isProviderEnabled('google')`).
- This needs no cloud-specific branch. Cloud is fixed to `gpt-5.6-luna`, so under this rule it simply never offers video.
- Keep the signature detection either way, so a clip named `image.jpg` is still identified rather than sent to the provider as a broken image.
- Add a degrade on the chat side, since the selection can change mid-thread: replace a file part the selected model cannot read with the text placeholder `capHistoricalAttachments` already uses (`describeAttachment`). That is what stops a model switch from reproducing this throw against an attachment already stored in the history.

**If no**, closing this PR is all that is needed. An MP4 cannot be uploaded today, so nothing else has to change.

## Minor

`components/file-upload-button.tsx` is not imported anywhere in the tree, so that part of this change is dead code. With it counted, the allowlist now lives in four places.

The two P2s from codex above (compatible brands, duration-independent token estimate) I agree with. They apply in the "yes" branch.

UI is not a blocker: video falls back to the same generic rendering PDF gets today, an extension chip in the composer, a link in the message, a generic icon in Library. Worth its own PR rather than this one.

miurla:
Thanks for the patch. Closing: only `google` can carry a video part today, so accepting MP4 at upload just moves the failure into the conversation. If video lands later, the gate belongs on the model's provider, not the upload boundary.

PR #985, opened 2026-08-21
fix(analytics): remove feedback control impression

### LINKED ISSUE #984: The feedback control impression event has outlived the question it was added to answer
## Problem

`feedback_control_shown` (`components/message-actions.tsx`) fires on every completed assistant answer that is presented. It was added to separate two explanations for sparse thumb feedback: users not clicking, versus the control never rendering at all. The thumbs render only inside `{traceId && ...}`, so a missing trace id removes the whole control silently, and the event carries `hasTraceId` precisely to make that case visible.

That question is settled. The delivery hypothesis did not survive: the control renders with a usable trace id, and since #981 the click path is instrumented directly on the same control (`feedback_control_clicked` / `feedback_recorded` / `feedback_failed`). Telling "never offered" from "offered and ignored" no longer requires an impression event, because the clicks are now counted where they happen.

What is left is a per-answer client event that keeps firing to supply a denominator for a rate that is no longer in question. Every other event in this file fires only on an explicit user action; this one fires on rendering.

## Evidence

Mechanism, from the code:

- `components/message-actions.tsx` reports the impression from an effect gated on `visible && !isLoading`, deduped per `messageId` in a module-scope `Set`. `visible` comes from `shouldShowActions = isLastTextPart && isStreamingComplete` in `components/render-message.tsx`, so one event means one completed answer presented. The dedup survives remounts within a page but not a reload, so revisiting a conversation adds impressions.
- Its properties are `hasTraceId`, `hasFeedback`, `chatId`, `isGuest`. `hasTraceId` exists only to test the delivery hypothesis. `hasFeedback` exists only so that already-scored controls, which are disabled and cannot produce another score, can be excluded from the denominator.
- The click side of the same control is covered by #981, in the same file.

Precedent: #896 removed `note_save_button_shown` from this same file for this same reason, an exposure event tied to message rendering that outlived its purpose.

How to re-check this before acting: compare the count of `feedback_control_clicked` against impressions restricted to `hasTraceId: true` and `hasFeedback: false` over the same window, and check whether the `hasTraceId: false` slice appears at all.

## Proposed direction

Make an explicit choice among three, rather than letting the event persist by default.

1. **Drop it**, following #896. Clicks then have to be read against a coarser exposure proxy such as answers rendered, and the offered-denominator is gone.
2. **Sample it.** Event volume falls proportionally and the event-rate denominator survives as a scaled estimate, but the distinct-user denominator degrades: at a low sampling rate a user with few answers in a day may not appear at all, so "share of exposed users who clicked" stops being directly readable. This is the real tension in this decision, and it is why sampling is not automatically the safe middle.
3. **Keep it, minus the dead properties.** Removing `hasTraceId` once the delivery hypothesis is closed reduces payload but not event count, so it addresses the smaller half of the cost.

The thing to weigh is whether the offered-denominator has continuing value: it is what would let a future change to the feedback control be judged on conversion rather than on raw click counts.

If it is kept in any form, the definition of `visible` is what would silently break it. An answer-rendering refactor that makes `shouldShowActions` hover-driven or viewport-driven would change what one impression means without touching this file.

## Acceptance criteria

- One decision is recorded, with its reason, instead of the event surviving by inertia.
- If sampled or dropped, any read that used the impression count as a denominator is either migrated to the new basis or explicitly retired, not left silently rescaled.
- If kept, the properties that no longer distinguish anything are removed.

Refs #968, #981, #896.

sushanttiwari-14:
## Problem

`feedback_control_shown` fires whenever a completed assistant answer is presented. It was introduced in #968 to determine whether sparse feedback resulted from users ignoring the control or from the control not rendering because a trace ID was missing.

That delivery question is now settled, and #981 instruments the interaction path directly through `feedback_control_clicked`, `feedback_recorded`, and `feedback_failed`. The impression event therefore continues generating one client event per presented answer without answering an active question.

## Decision

Drop the impression event, following the precedent established in #896.

This explicitly retires the offered-control denominator. Feedback usage should now be read from direct click and outcome events rather than treating `feedback_control_shown` as a continuing exposure metric. There are no checked-in dashboards, queries, or other consumers that require migration.

## Changes

- Remove the `feedback_control_shown` effect and its module-level deduplication set.
- Remove tests covering the retired impression behavior.
- Add a regression test confirming that rendering the feedback control does not emit an analytics event.
- Keep `feedback_control_clicked`, `feedback_recorded`, and `feedback_failed` unchanged.

## Verification

- `bun lint`
- `bun typecheck`
- `bun format:check`
- `bun run test` — 560 passed, 3 skipped
- `bun run build`

Closes #984

### DIFF
--- a/components/__tests__/message-actions.test.tsx
+++ b/components/__tests__/message-actions.test.tsx
@@ -1,5 +1,3 @@
-import React from 'react'
-
 import { fireEvent, render, screen, waitFor } from '@testing-library/react'
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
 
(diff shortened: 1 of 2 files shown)


miurla:
LGTM :+1:

PR #989, opened 2026-08-22
fix(search): fall back after recoverable provider failures

### LINKED ISSUE #980: A rate-limited general search provider fails the search tool call instead of falling back
## Problem

When a dedicated general search provider is configured and it rate-limits, the whole search tool
call fails. There is no runtime fallback to the provider that would have served the query if the
dedicated one had never been configured.

The fallback that does exist is configuration-time only. `type="general"` uses the dedicated general
provider when one is available and otherwise falls back to the optimized provider. Once the
dedicated provider is configured, that branch is settled before the request runs, so a provider that
is present but failing has no path back to it.

## Evidence

Mechanism, from the code:

- `lib/tools/search.ts` resolves the provider once: `getGeneralSearchProviderType()` for
  `type="general"`, with the optimized provider used only when no dedicated general provider is
  available. The log line on that branch says it plainly: "no dedicated provider available, using
  optimized search provider".
- `createSearchProvider` (`lib/tools/search/providers/index.ts`) is a static switch. It has no
  notion of a provider being temporarily unusable.
- In `lib/tools/search/providers/brave.ts`, `search()` runs the requested content types with
  `Promise.allSettled` and rethrows the first rejection when *every* requested type failed. A rate
  limit that hits all requested content types is therefore a thrown failure.
- Back in `lib/tools/search.ts`, the `catch` wraps it in `ToolFailureError` and rethrows, so the
  model gets a failed search rather than results from another provider.

Both the web and the video endpoints of the configured general provider have produced HTTP 429 in
production, which is the shape that reaches this path.

Note the partial case already behaves correctly: when one content type is rate-limited and another
succeeds, `search()` returns what it got. The gap is specific to a request where every requested
content type fails.

Read `search` observations at `level: ERROR` and look for status messages naming the provider with
an HTTP 429. #964 is what makes these legible, and #961 is what puts the status code in them.

## Proposed direction

Decide whether a general-search failure should degrade to the optimized provider at request time
rather than fail the turn, and if so, which failures qualify. A rate limit and a transport error are
recoverable in a way that a malformed query is not, so the trigger should be narrow rather than a
blanket catch.

Two smaller judgment calls sit inside that:

- Whether a fallback result should be marked as such, given that the two providers do not return the
  same shape and the multimedia content types may simply be absent from the fallback.
- Whether repeated rate limiting should suppress the dedicated provider for a short window instead of
  paying a failed request every time. That is more machinery than the problem may warrant, and it
  should be an explicit choice rather than a side effect.

This is filed rather than fixed because the answer is a product call about what a degraded search is
allowed to look like, not just a code path.

## Acceptance criteria

- A rate-limited general search provider does not, by itself, fail the search tool call when another
  configured provider could have answered.
- The fallback triggers only on failures that another provider could plausibly recover from.
- A search served by the fallback is distinguishable in the trace from one served by the configured
  provider.

Refs #961, #963, #964.

sushanttiwari-14:
## Summary

- Fall back to the configured optimized search provider when the dedicated general search provider encounters a recoverable failure.
- Limit fallback behavior to HTTP 408, HTTP 429, HTTP 5xx, and recognized transport failures.
- Preserve existing behavior for client errors and partial multimedia results.
- Avoid retrying when the dedicated and optimized providers are the same.
- Log the provider transition and failure category so fallback requests are distinguishable in traces.

## Testing

- Added coverage for rate-limit fallback.
- Added coverage for transport-error fallback.
- Verified that client errors do not trigger fallback.
- Verified that the same provider is not retried as its own fallback.
- Verified recoverable HTTP status and network-error classification.

## Validation

- `bun run test` — 586 tests passed
- `bun lint`
- `bun typecheck`
- `bun format:check`
- `bun run build`

Closes #980

### DIFF
--- a/lib/tools/__tests__/search-fallback.test.ts
+++ b/lib/tools/__tests__/search-fallback.test.ts
@@ -1,0 +1,127 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const mocks = vi.hoisted(() => ({
+  braveSearch: vi.fn(),
+  createSearchProvider: vi.fn(),
+  tavilySearch: vi.fn()
+}))
+
+vi.mock('@/lib/tools/search/providers', () => ({
+  createSearchProvider: mocks.createSearchProvider,
+  DEFAULT_PROVIDER: 'tavily'
+}))
+
+import { createSearchTool } from '@/lib/tools/search'
+
+const fallbackResult = {
+  results: [
+    {
+      title: 'Fallback result',
+      content: 'Fallback content',
+      url: 'https://example.com/result'
+    }
+  ],
+  images: [],
+  query: 'current events',
+  number_of_results: 1
+}
+
+function executeGeneralSearch() {
+  const result = createSearchTool('openai:gpt-4o-mini').execute?.(
+    {
+      query: 'current events',
+      type: 'general',
+      content_types: ['web'],
+      max_results: 10,
+      search_depth: 'basic',
+      include_domains: [],
+      exclude_domains: []
+    },
+    { toolCallId: 'search-call', messages: [], context: {} }
+  )
+
+  return (result as AsyncIterable<unknown>)[Symbol.asyncIterator]()
+}
+
+describe('general search provider fallback', () => {
+  beforeEach(() => {
+    vi.stubEnv('BRAVE_SEARCH_API_KEY', 'test-key')
+    vi.stubEnv('SEARCH_API', 'tavily')
+    mocks.createSearchProvider.mockImplementation(provider => ({
+      search: provider === 'brave' ? mocks.braveSearch : mocks.tavilySearch
+    }))
+    mocks.tavilySearch.mockResolvedValue(fallbackResult)
+    vi.spyOn(console, 'log').mockImplementation(() => undefined)
+    vi.spyOn(console, 'error').mockImplementation(() => undefined)
+  })
+
+  afterEach(() => {
+    vi.unstubAllEnvs()
+    vi.restoreAllMocks()
+    vi.clearAllMocks()
+  })
+
+  it('uses the optimized provider after the general provider is rate limited', async () => {
+    mocks.braveSearch.mockRejectedValue(
+      Object.assign(new Error('Brave search failed: HTTP 429'), { status: 429 })
+    )
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => undefined)
+    const iterator = executeGeneralSearch()
+
+    await iterator.next()
+    const complete = await iterator.next()
+
+    expect(complete.value).toMatchObject({
+      state: 'complete',
+      results: fallbackResult.results,
+      toolCallId: 'search-call'
+    })
+    expect(mocks.braveSearch).toHaveBeenCalledOnce()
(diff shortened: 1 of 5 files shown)


miurla (changes requested):
Thanks for taking this on. The fallback design is right: the trigger is narrow, the same provider is never retried as its own fallback, and the recoverable classifier walks the cause chain instead of matching on message text alone. The Brave path this targets throws only when every requested content type fails, and that error carries a numeric `status` from `BaseSearchProvider.createHttpError`, so `isRecoverableSearchError` sees it as intended.

One blocking issue, and it is unrelated to the fallback itself.

### The `|| 'basic'` default is dropped

Before:

```ts
const effectiveSearchDepthForAPI =
  searchAPI === 'searxng' && process.env.SEARXNG_DEFAULT_DEPTH === 'advanced'
    ? 'advanced'
    : effectiveSearchDepth || 'basic'
```

After, in `getEffectiveSearchDepth`, the `|| 'basic'` is gone and `requestedDepth` is returned as is.

That guard is not redundant. In the standard schema (`lib/schema/search.tsx`), `search_depth` is `z.string().optional().default('basic')`, a free-form string rather than an enum, and a zod default only applies to `undefined`. A model that emits `"search_depth": ""` therefore passes validation, and `''` is falsy, so the old expression turned it into `'basic'` while the new one forwards it. The strict schema uses `z.enum`, so this only affects the non-strict path, which is every model except OpenAI `o*`.

Downstream, `TavilySearchProvider.search` declares `searchDepth: 'basic' | 'advanced' = 'basic'`, and a default parameter also only applies to `undefined`, so `''` is not corrected there either. It goes into the request body as `search_depth: ""`, which the API rejects. The net effect is that a search that used to succeed becomes a failed tool call.

Type checking cannot catch it, because `effectiveSearchDepth` reaches the helper through the existing `search_depth as 'basic' | 'advanced'` cast.

Reproduced by applying this branch on top of `main` and running the same test against both trees: it passes on `main` and fails here.

```ts
// executes the tool with search_depth: '' and asserts on what the provider receives
const depth = tavilySearch.mock.calls[0][2]
expect(depth).toBe('basic')   // main: 'basic'   this branch: ''
```

Restoring the default inside the helper is enough:

```ts
return provider === 'searxng' && process.env.SEARXNG_DEFAULT_DEPTH === 'advanced'
  ? 'advanced'
  : requestedDepth || 'basic'
```

A test covering an empty `search_depth` would be worth adding alongside it, since nothing in the suite currently pins that behaviour.

### Two non-blocking notes

- `describeRecoverableSearchError` reads `status` only from the top-level error, while `isRecoverableSearchError` walks the `cause` chain. An error whose status sits on a nested cause is classified as recoverable but logged as `transport error`. Log-only, but the two functions should agree on where they look.
- The fallback is currently recorded with `console.warn` only. Nothing on the tool result or the span marks the request as fallback-served, so a fallback search is distinguishable in the runtime logs but not in the trace itself. That is the third acceptance criterion in #980, so it may be worth carrying the provider through on the result rather than closing that criterion on the log line.

sushanttiwari-14:
Thanks for the detailed review. I’ll restore the `|| 'basic'` fallback, add a regression test, and address the two trace related notes. I’ll push the updates after running all checks.

miurla (approved):
:+1:

PR #997, opened 2026-08-25
feat: add Serping API search provider

musabgultekin:
This adds [serpingapi.com](https://serpingapi.com) as a search provider option (`SEARCH_API=serpingapi`).

**Disclosure: I run this service.** It's a small SERP API (real-time Google results as JSON, one POST endpoint); the free tier is 250 searches/month with no credit card, which makes it an easy default for self-hosters trying out Morphic. Happy to adjust anything to fit the project's conventions.

### Changes

- `lib/tools/search/providers/serpingapi.ts` — new provider extending `BaseSearchProvider`, modeled on the Tavily/Brave providers: POST to `/v1/search` with `X-API-Key` auth, maps `organic[]` (`title`/`link`/`snippet`) into `SearchResults` items (`title`/`url`/`content`)
- `lib/tools/search/providers/index.ts` — adds `serpingapi` to `SearchProviderType` and the factory
- `.env.local.example` — documents `SERPINGAPI_API_KEY` and adds `serpingapi` to the `SEARCH_API` options
- `docs/CONFIGURATION.md` — short section under Search Providers

### Notes

- Like Firecrawl, the API doesn't support `includeDomains`/`excludeDomains`, so those are accepted but not forwarded (noted in a comment).
- It returns web results only; `images` is returned empty and video/image `content_types` remain the domain of the dedicated general provider (Brave), unchanged.
- Errors go through `createHttpError` so the recoverable-error fallback classification works as with the other fetch-based providers.

`bun lint`, `bun typecheck`, `bun format:check`, `bun run test` (614 passed) and `bun run build` all pass.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -74,6 +74,10 @@ TAVILY_API_KEY=[YOUR_TAVILY_API_KEY]
 # Get your API key at: https://www.firecrawl.dev/app
 # FIRECRAWL_API_KEY=[YOUR_FIRECRAWL_API_KEY]
 
+# Option 4: Serping API (Real-time Google search results)
+# Get your API key at: https://serpingapi.com/
+# SERPINGAPI_API_KEY=[YOUR_SERPINGAPI_API_KEY]
+
 # Brave Search (Optional - video/image support for general searches)
 # Get your API key at: https://brave.com/search/api/
 # BRAVE_SEARCH_API_KEY=[YOUR_BRAVE_SEARCH_API_KEY]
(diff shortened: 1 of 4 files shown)


musabgultekin:
Verified this provider against the live API today. The exact request it sends — `POST https://api.serpingapi.com/v1/search` with `X-API-Key` and body `{"q":"...","num":10}` — returns 200 with `organic[]` entries carrying `title`, `link`, and `snippet`, which map cleanly to `results[].title/url/content`. Invalid keys return a 401 JSON error, so the error path works as written too. The earlier upstream outage is resolved.

miurla:
Thanks for the PR, and for the upfront disclosure. The implementation is clean and follows the existing provider conventions.

We're intentionally keeping the set of built-in search providers small and aren't adding new provider implementations right now. This is the same call we've made on recent provider contributions, so it's about maintenance surface, not Serping specifically.

Closing for that reason. Appreciate the contribution.

PR #1000, opened 2026-08-26
fix(errors): classify deterministic preparation failures

### LINKED ISSUE #979: Deterministic preparation failures are reported to the user as retryable unknown errors
## Problem

A turn that fails during the preparation path, before the model is ever called, is presented to the
user as a generic retryable failure: "We could not generate a response. Please try again."

Several preparation failures are deterministic. The same request, replayed with the same stored
state, fails in exactly the same place. Telling the user to try again is then advice that cannot
work, and it costs them a full retry to find that out.

## Evidence

Mechanism, all visible in the code:

- `prepareMessages` (`lib/streaming/helpers/prepare-messages.ts`) raises bare `new Error(...)` for
  precondition failures on the regenerate path: the chat carries no messages, or the target message
  cannot be found and no fallback position exists. Neither condition can resolve itself on a retry.
- The DB row to `UIMessage` mapping (`lib/utils/message-mapping.ts`), which `loadChat` goes through
  on that same path, raises bare `new Error(...)` for a persisted part it cannot map: an unknown
  part type, or a tool part with no `tool_state`. These describe stored data, so they reproduce on
  every attempt.
- All of these reach `createPublicErrorResponse` in the two stream response builders. A bare `Error`
  carries no status and no `code`, so `classifyError` in `lib/errors/public-error.ts` falls through
  every branch to `code: 'unknown'`, `retryable: true` and the default message.

The module already has the vocabulary for the opposite case. `malformed_request` pairs "This
conversation could not be sent to the model. Please start a new chat." with the detail "Retrying
this message will fail the same way." Today nothing in the preparation path can reach it, because it
is gated on a provider status code plus a provider signature.

There is a second cost. The `unknown:` bucket is what #954, #956 and #962 use to find genuine
unclassified defects. Known preconditions landing in it dilute exactly the signal those changes were
built to sharpen.

Read `research` observations at `level: ERROR` whose `statusMessage` starts `unknown:` and whose
`metadata.streamErrorPhase` is `preparation`, then check `metadata.streamErrorShape`. A shape that
is a bare error name with no `code` is one of these. #978 adds `metadata.streamErrorStage`, which
names the step it came from.

## Proposed direction

1. Give the deterministic preparation preconditions an explicit identity instead of a bare `Error`,
   so they classify as themselves rather than as `unknown`. Which conditions qualify is the judgment
   call here: a missing chat or an unmappable stored part is deterministic, while a database timeout
   in the same call is not, and only the first group should tell the user that retrying is futile.
2. Decide whether these reuse `malformed_request` or get their own code. The existing copy fits the
   regenerate cases well, but "start a new chat" is not obviously the right instruction for every
   condition, and the wording is user-facing.
3. Honour an explicit code only when this codebase produced the error, following the provenance
   discipline already written into `isToolFailurePayload` and `isMalformedRequestPayload`. An
   arbitrary `code` property on a third-party error must not be able to select a public code.

Item 1 is worth sequencing after #978 has been read in production, since the stage marker will show
which of these sites actually fire.

## Acceptance criteria

- A deterministic preparation failure is not reported to the user as retryable.
- The same failure classifies as something other than `unknown`, so the unclassified bucket keeps
  meaning what #956 and #962 need it to mean.
- A failure in the same path that genuinely might succeed on retry keeps its current treatment.
- No user-supplied content is added to any trace or to any client payload by the change.

Refs #954, #956, #962, #978.

sushanttiwari-14:
## Summary

Fixes deterministic preparation failures being classified as retryable `unknown` errors.

Preparation failures caused by invalid stored conversation state now use an explicit internal error type and are presented as non-retryable `malformed_request` errors. Transient failures, such as database timeouts, retain their existing retryable behavior.

## Changes

- Add `DeterministicPreparationError` to identify failures that cannot succeed when replayed with the same state.
- Use it for:
  - Regeneration with no stored messages.
  - Regeneration with no matching or fallback message.
  - Persisted message parts that cannot be mapped.
  - Persisted tool parts with missing or invalid state.
- Map these failures to the existing sanitized `malformed_request` response.
- Preserve provenance by recognizing only errors created by the application, rather than trusting arbitrary error properties.
- Add tests covering classification, serialization, sanitization, persisted-part mapping, and transient failures.

## Testing

- `bun lint`
- `bun typecheck`
- `bun format:check`
- `bun run test`
  - 618 passed
  - 3 skipped
- `bun run build`

Fixes #979

### DIFF
--- a/lib/errors/__tests__/public-error.test.ts
+++ b/lib/errors/__tests__/public-error.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
+import { DeterministicPreparationError } from '@/lib/errors/deterministic-preparation-error'
 import {
   createPublicErrorResponse,
   getPublicRateLimitDetails,
@@ -220,6 +221,45 @@ describe('public error mapping', () => {
       'Retrying this message will fail the same way.'
     )
     expect(payload.retryable).toBe(false)
+  })
+
+  it('maps deterministic preparation failures to non-retryable copy', () => {
+    const payload = toPublicErrorPayload(
+      new DeterministicPreparationError(
+        'Unknown part type: private-user-content'
+      )
+    )
+
+    expect(payload.code).toBe('malformed_request')
+    expect(payload.error).toBe(
+      'This conversation could not be sent to the model. Please start a new chat.'
+    )
+    expect(payload.details).toBe(
+      'Retrying this message will fail the same way.'
+    )
+    expect(payload.retryable).toBe(false)
+    expect(JSON.stringify(payload)).not.toContain('private-user-content')
+  })
+
+  it('preserves deterministic preparation copy through serialization', () => {
+    const serialized = serializePublicError(
+      new DeterministicPreparationError('No messages found')
+    )
+    const payload = toPublicErrorPayload(new Error(serialized))
+
+    expect(payload.code).toBe('malformed_request')
+    expect(payload.retryable).toBe(false)
+  })
+
+  it('does not trust a foreign deterministic preparation error name', () => {
+    const payload = toPublicErrorPayload({
+      name: 'DeterministicPreparationError',
+      message: 'Unknown part type: private-user-content'
+    })
+
+    expect(payload.code).toBe('unknown')
+    expect(payload.retryable).toBe(true)
+    expect(payload.error).not.toContain('private-user-content')
   })
 
   it('does not honour a malformed_request code carrying foreign copy', () => {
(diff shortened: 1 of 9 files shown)


miurla (approved):
Thanks for the follow-up. The initial chat load is now inside the same public-error boundary, so deterministic preparation failures reach the client as non-retryable `malformed_request` instead of a retryable unknown error. Transient failures (DB/Postgres timeouts) still classify as retryable, so the existing behavior is preserved.

Codex's P2 finding is addressed and its re-review came back clean. CI is green on 56ffd99.

LGTM 👍

PR #1007, opened 2026-09-02
fix(citations): cite searches by a short server-assigned id

miurla:
## Problem

Both mode prompts tell the model to cite as `[number](#toolCallId)`, so writing a citation requires reproducing a provider tool call id verbatim. The configured model often does not: it alters the id, drops or adds characters, or writes an id in a shape it has never been given. `processCitations` cannot resolve those, and the citation is replaced with an empty string, so the answer silently loses the source it was pointing at.

## Root cause

The citation key is an identifier chosen by the provider for its own bookkeeping. It is long, opaque, and carries a prefix the prompt then has to spend rules telling the model not to touch. Nothing in the pipeline gives the model an identifier that is cheap to copy.

## Fix

- `lib/tools/search.ts` attaches a short `citeId` to the search output, derived deterministically from the tool call id so it can be recomputed from output persisted before this change.
- The model view built by `toModelOutput` leads with `citeId` and carries only that id when it is present. Two ids on one result invite the model to cite the wrong one; older output with no `citeId` still carries `toolCallId`.
- `lib/utils/citation.ts` keys each citation map by both id forms and exposes one `resolveCitationMap` used by both call sites, so answers written before this change keep resolving. The lookup prefers an exact match and only then normalizes, and a short id that would collide with one already claimed is not registered, so a citation is dropped rather than resolved to the wrong source.
- `lib/streaming/helpers/compact-historical-messages.ts` drops its duplicate lookup and uses the shared one.
- Both citation sections of `lib/agents/prompts/search-mode-prompts.ts` now name the `citeId` and show it in the examples.

No component, schema or migration is touched. Rendering is unchanged: it happens after citations are substituted with `[domain](url)`.

## Verification

- `bun lint`, `bun typecheck`, `bun format:check`, `bun run test` all pass.
- New tests cover the derivation, resolution of an explicit short id, resolution of a legacy long id, resolution of a legacy output by its derived short id, an unknown id resolving to nothing, cited-source extraction in history for both id forms, and the model view keeping exactly one id.

## Status

Draft. The change is measured offline against an internal evaluation set, and on that harness citation resolution did not improve; the failures that remain are answers where the model writes an identifier it was never given, in a shape of its own, rather than mangling the one it was handed. Shortening the identifier does not remove that behavior, which suggests the remaining gap is the two-coordinate citation format itself rather than the length of the id, and points at a per-result identifier as the next thing to try. Opening for review of the mechanism and the backward-compatibility path; not for merge on the current evidence.

### DIFF
--- a/lib/agents/prompts/search-mode-prompts.ts
+++ b/lib/agents/prompts/search-mode-prompts.ts
@@ -61,7 +61,7 @@ Your approach:
    - Use concrete examples and specific data when available
    - Avoid unnecessary elaboration while maintaining clarity
    - Scale response length naturally based on query complexity
-5. **CRITICAL: You MUST cite sources inline using the [number](#toolCallId) format**
+5. **CRITICAL: You MUST cite sources inline using the [number](#citeId) format**
 
 Tool preamble (keep very brief):
 - Start directly with search tool without text preamble for efficiency
@@ -81,7 +81,7 @@ Search requirement (MANDATORY):
 - Do NOT answer informational questions based only on internal knowledge; verify with current sources via search and cite
 - Prefer recent sources when recency matters; mention dates when relevant
  - For informational questions without URLs, your FIRST action in this turn MUST be the \`search\` tool. Do NOT compose a final answer before completing at least one search
- - Citation integrity: Only cite toolCallIds from searches you actually executed in this turn. Never fabricate or reuse IDs
+ - Citation integrity: Only cite citeIds from search responses for searches you actually executed in this turn. Never fabricate or reuse IDs
  - If initial results are insufficient or stale, state the limitation or ask a clarifying question; do not run a second search
 
 Fetch tool usage:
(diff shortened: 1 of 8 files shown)


miurla:
Closing. The premise did not survive measurement.

This change was built on the expectation that citations fail because the model cannot reproduce a long provider tool call id, so a short id would be copied reliably. Measured offline against an internal evaluation set, citation resolution did not improve: it moved the wrong way, while the share of answers that carry citations at all stayed flat. The failures are not mangled ids. The model writes an identifier it was never given, in a shape of its own that encodes the position of the result it is citing, and it does that more readily against a short id than against the long one it was previously handed.

That points at the citation format rather than the identifier. A citation is currently two coordinates, a search and an offset within it, and the invented ids look like an attempt to collapse those into one. The thing worth trying next is a per-result identifier, which is a different change to a different surface, not an iteration on this diff.

The branch is kept for the parts that would carry over: the shared citation lookup and the backward-compatible resolution of both id forms.

PR #1012, opened 2026-09-07
feat(library): move the Library entry point into the sidebar

miurla:
## Problem

The Library entry point lives in the header action row (`components/header.tsx`), next to Feedback and the user menu. Library is navigation, not a per-page action, so it competes for the header's always-visible slots with the controls that actually belong there.

There is also a real inconsistency underneath it. `openLibrary` / `toggleLibrary` in `components/library/library-context.tsx` call `setOpen(false)` to get the sidebar out of the way, but `setOpen` in `components/ui/sidebar.tsx` only touches the desktop `open` state. On mobile the sidebar is a `Sheet` driven by `openMobile`, which that call never affects. With the trigger in the header (outside the sidebar) this is invisible, because the sheet is already closed. Moving the entry point inside the sidebar surfaces it: the sheet would stay open with the library panel underneath it.

Sidebar usage is also unmeasured. `autocapture` is disabled (`lib/analytics/posthog-client.ts`), and nothing emits an event when the sidebar is opened or closed, so after this move there is no way to tell "the entry point is not being found" apart from "it is found and not used".

## Root cause

- The header is the only entry point to the library, so navigation and actions share one row.
- `setOpen` and `setOpenMobile` are separate states in `SidebarProvider`, and the library context only ever calls the first one.
- `toggleSidebar` has no instrumentation.

## Fix

- New `components/sidebar/library-menu-item.tsx` renders the Library entry as a `SidebarMenuItem`, placed directly under `NewChatMenuItem` in `components/app-sidebar.tsx`. `app/layout.tsx` already gates `AppSidebar` on a signed-in user and nests `LibraryProvider` inside `SidebarProvider`, so no extra guard or wiring is needed.
- Remove the Library button and its `useLibrary` / icon imports from `components/header.tsx`. The entry point lands on one surface rather than two, so the `source` property stays readable.
- `library-context.tsx` now branches on `isMobile` and closes the surface that is actually open: `setOpenMobile(false)` on mobile, `setOpen(false)` on desktop.
- `toggleSidebar` in `components/ui/sidebar.tsx` emits `sidebar_toggled` with the new open value and `isMobile`. It is placed on the provider's toggle rather than on `SidebarTrigger` so every path that toggles the sidebar is counted, including the keyboard shortcut.

## Verification

- `components/sidebar/__tests__/library-menu-item.test.tsx`: clicking the item calls `toggleLibrary` and records `source: 'sidebar'`, with the open and closed event names selected from the current state, and the toggle running before the capture.
- `components/library/__tests__/library-context.test.tsx`: opening or toggling the library closes the mobile sheet on mobile and the desktop sidebar on desktop, and never the other one; toggling the library closed leaves both alone.
- `components/ui/__tests__/sidebar.test.tsx`: `sidebar_toggled` reports the new open value on both desktop and mobile, across an open and a close.
- `bun lint`, `bun typecheck`, `bun format:check`, `bun run test`, `bun run build` all pass.

### DIFF
--- a/components/__tests__/app-sidebar.test.tsx
+++ b/components/__tests__/app-sidebar.test.tsx
@@ -1,0 +1,55 @@
+import { render, screen } from '@testing-library/react'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+
+import { SidebarProvider } from '@/components/ui/sidebar'
+
+import AppSidebar from '../app-sidebar'
+
+vi.mock('../sidebar/chat-history-section', () => ({
+  ChatHistorySection: () => <div />
+}))
+
+vi.mock('../sidebar/chat-history-skeleton', () => ({
+  ChatHistorySkeleton: () => <div />
+}))
+
+vi.mock('../sidebar/new-chat-menu-item', () => ({
+  NewChatMenuItem: () => <div>New</div>
+}))
+
+vi.mock('../sidebar/library-menu-item', () => ({
+  LibraryMenuItem: () => <div>Library</div>
+}))
+
+vi.mock('@/hooks/use-mobile', () => ({
+  useIsMobile: () => false
+}))
+
+function renderSidebar() {
+  render(
+    <SidebarProvider>
+      <AppSidebar />
+    </SidebarProvider>
+  )
+}
+
+describe('AppSidebar library entry', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs()
+  })
+
+  test('shows the Library entry when authentication is enabled', () => {
+    vi.stubEnv('ENABLE_AUTH', 'true')
+    renderSidebar()
+
+    expect(screen.getByText('Library')).toBeInTheDocument()
+  })
+
+  test('hides the Library entry in anonymous mode', () => {
+    vi.stubEnv('ENABLE_AUTH', 'false')
+    renderSidebar()
+
+    expect(screen.queryByText('Library')).not.toBeInTheDocument()
+    expect(screen.getByText('New')).toBeInTheDocument()
+  })
+})
--- a/components/app-sidebar.tsx
+++ b/components/app-sidebar.tsx
@@ -14,6 +14,7 @@ import {
 
 import { ChatHistorySection } from './sidebar/chat-history-section'
 import { ChatHistorySkeleton } from './sidebar/chat-history-skeleton'
+import { LibraryMenuItem } from './sidebar/library-menu-item'
 import { NewChatMenuItem } from './sidebar/new-chat-menu-item'
 import { IconLogo } from './ui/icons'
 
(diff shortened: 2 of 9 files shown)


miurla:
Confirmed and fixed in 7fda115.

The asymmetry is exactly as described: `app/layout.tsx` derives `user` from Supabase and `userId` from `getCurrentUserId()`, so with `ENABLE_AUTH=false` the anonymous id makes `{userId && <AppSidebar />}` render while `user` stays null. The header button was gated on `user &&`, so it was hidden in that mode, and the unconditional sidebar item was not. Clicking it would have opened a panel whose note and file loads are rejected by the guards in `lib/actions/notes.ts` and `lib/actions/files.ts`.

`AppSidebar` is a server component, so it now reads the same switch the route pages already use for `libraryAvailable` (`process.env.ENABLE_AUTH !== 'false'`) and omits the item in anonymous mode. Added `components/__tests__/app-sidebar.test.tsx` covering both directions.

miurla:
Both confirmed and fixed in e4a5fac.

**Stale desktop state on mobile.** Verified: `components/artifact/artifact-context.tsx` reads `open` (the desktop value) as `sidebarOpen` and runs `if (sidebarOpen && libraryOpen) closeLibrary()`. Before this PR `toggleLibrary` always called `setOpen(false)`, so that invariant held; the mobile branch I added skipped it, which is what made a stale `open` reachable. `openLibrary` and `toggleLibrary` now go through one `closeSidebarSurfaces` helper that always clears the desktop value and additionally clears `openMobile` on mobile.

**Mobile dismissals bypassing the capture.** Also correct: `Sheet open={openMobile} onOpenChange={setOpenMobile}` means Escape and overlay clicks set the state directly, so instrumenting `toggleSidebar` alone would have recorded opens without their matching closes. Rather than instrument a second path, the capture now lives on the state transitions themselves: `setOpen` and a wrapped `setOpenMobile` each emit `sidebar_toggled` when the value actually changes, and `toggleSidebar` is back to its original form. Every route into the state is covered, including the Sheet's own dismissal and the programmatic close when Library opens, and a set that changes nothing emits nothing.

Tests added for the mobile dismissal path, for the no-op guard, and for both surfaces being cleared when Library opens.

miurla:
Confirmed and fixed in ff409ac. This one was created by the previous two fixes interacting: clearing both surfaces plus capturing on both setters meant a mobile Library click emitted a desktop-labelled close for a sidebar that was never on screen.

The capture is now conditioned on the active surface: `setOpen` records only when the viewport is not mobile, and `setOpenMobile` only when it is. Clearing the hidden desktop value on a narrow viewport stays silent, so no event describes a surface the user could not see, and the surviving events are one per real transition. Added a test for the mobile clear emitting nothing.

miurla:
Confirmed and fixed in 42ab530. This is the mirror of the earlier finding, and it is only safe to fix now because the capture is gated on the visible surface: clearing a hidden value no longer invents an event, so both surfaces can be cleared unconditionally.

`closeSidebarSurfaces` now calls `setOpen(false)` and `setOpenMobile(false)` with no viewport branch, and `isMobile` is no longer needed in `LibraryProvider`. Whichever surface the viewport switches to next is already closed, so neither the sheet nor the desktop sidebar can come back on top of an open Library panel. The library-context test now asserts both setters are called in both viewport cases.

PR #1013, opened 2026-09-07
fix(citation): stop a malformed closing delimiter from swallowing answer text

miurla:
## Problem

A citation whose closing delimiter is `]` instead of `)` (`[1](#S37]`) does not just fail to render. It deletes a span of the answer.

`processCitations` in `lib/utils/citation.ts` matched citations with `/\[\s*(\d+)\s*\]\(#([^)]+)\)/g`. The id class `[^)]+` does not stop at the malformed `]`, so it keeps consuming until the next `)` anywhere later in the text, which in a cited answer is normally the close of a later, well-formed citation. The whole run becomes one match with an id that resolves to nothing, and the callback returns `''` for an unresolvable id, so everything between the two citations is replaced by nothing.

Reproduction, on the pattern alone:

```
input: 'Alpha is true. [1](#S37] Beta is a whole sentence that should survive. Gamma cites correctly. [1](#S5)'

single match, captured id:
  'S37] Beta is a whole sentence that should survive. Gamma cites correctly. [1](#S5'

processCitations output:
  'Alpha is true. '
```

Three surfaces are affected, because they all go through this function or the copy of the pattern next to it:

- the rendered answer (`components/message.tsx`)
- the copied markdown (`components/message-actions.tsx`)
- the history replayed to the model. `compact-historical-messages.ts` runs historical assistant text through `processCitations`, and its own `CITATION_PATTERN` was a second copy of the same regex. `getCitedSources` therefore also loses the source excerpts for every valid citation inside the swallowed span, so the evidence for those citations stops being attached to the compacted turn.

The last one is the reason this is worth fixing beyond the visible artifact: a single malformed character removes text and evidence from what later turns are built on, silently.

## Root cause

The citation id character class was unbounded on everything except `)`. An id can never legitimately contain `]`, `[`, `(` or whitespace, but the pattern allowed all of them, so a missing `)` turned the id into a wildcard across the rest of the paragraph.

## Fix

- Bound the id class to `[^)\]\s]+`, so a match cannot cross a bracket or whitespace into neighbouring text. This alone contains the damage: a malformed citation can no longer reach past itself.
- Accept `]` as a closing delimiter as well as `)`, so a citation with a malformed close resolves to its source instead of being dropped or left as a literal string. The id still has to resolve against the citation maps, so this widens the delimiter, not what counts as a valid id.
- Define the pattern once, in `createCitationPattern()` in `lib/utils/citation.ts`, and have `compact-historical-messages.ts` import it instead of keeping its own copy, so the two readers cannot drift apart again. It is a factory rather than a shared constant because a `/g` RegExp carries `lastIndex`, and a single instance shared across call sites would skip matches.

No change to `resolveCitation`, `extractCitationMaps`, the label scheme, or any prompt.

## Verification

`lib/utils/__tests__/citation.test.ts`:

- a malformed close between two sentences no longer deletes the text after it, and the following well-formed citation still resolves
- a malformed close resolves to the same source as the same citation written correctly
- a well-formed citation resolves exactly as before
- an unresolvable id still yields an empty replacement and leaves the surrounding text alone

`lib/streaming/__tests__/compact-historical-messages.test.ts`:

- a source cited after a malformed-close citation still reaches the compacted turn's source context, which is the case that regressed silently

`bun lint`, `bun typecheck`, `bun format:check`, `bun run test`, `bun run build` all pass.

### DIFF
--- a/lib/streaming/__tests__/compact-historical-messages.test.ts
+++ b/lib/streaming/__tests__/compact-historical-messages.test.ts
@@ -215,6 +215,50 @@ describe('compactHistoricalMessages', () => {
     expect(sourceContext.text).not.toContain('x'.repeat(401))
   })
 
+  it('keeps a source cited after a malformed-close citation', () => {
+    const messages = [
+      {
+        id: 'assistant-1',
+        role: 'assistant',
+        parts: [
+          {
+            type: 'tool-search',
+            toolCallId: 'call_1',
+            state: 'output-available',
+            input: { query: 'example' },
+            output: {
+              query: 'example',
+              images: [],
+              results: [
+                {
+                  title: 'First source',
+                  url: 'https://example.com/first',
+                  content: 'First evidence'
+                },
+                {
+                  title: 'Later source',
+                  url: 'https://example.com/later',
+                  content: 'Later evidence'
+                }
+              ]
+            }
+          },
+          {
+            type: 'text',
+            text: 'First [1](#call_1] middle survives. Later [2](#call_1)'
+          }
+        ]
+      }
+    ] as unknown as UIMessage[]
+
+    const [, sourceContext] = compactHistoricalMessages(messages)[0]
+      .parts as Array<{ type: 'text'; text: string }>
+
+    expect(sourceContext.text).toContain('First source')
+    expect(sourceContext.text).toContain('Later source')
+    expect(sourceContext.text).toContain('https://example.com/later')
+  })
+
   it('uses description from persisted Brave results when content is absent', () => {
     const messages = [
       {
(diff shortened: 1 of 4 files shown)


miurla:
Confirmed and fixed in 6b639c6. Good catch: the comment claimed containment the character class did not actually provide.

Reproduced on the pattern alone with `Alpha. [1](#missing[1](#S5) tail.`. The previous class `[^)\]\s]+` stopped at the `]` of the following citation's opener, so the match ran to `[1](#missing[1]`, the unresolvable id was replaced with an empty string, and the valid `S5` citation lost its opener and never resolved. One malformed close therefore still took a valid citation with it, which is the failure this PR was meant to remove.

The id class is now `[^\s()[\]]+`, which excludes whitespace and all four bracket characters, so a match cannot cross into an adjacent citation in either direction. With that, the same input leaves `[1](#missing` as literal text and `[1](#S5)` resolves normally. Added a regression test for it in `lib/utils/__tests__/citation.test.ts`.

PR #1017, opened 2026-09-11
fix(search): require quick mode search before early stop

### LINKED ISSUE #1014: Quick mode's early-stop criterion lets the model skip the mandatory search
## Problem

The quick-mode system prompt states two rules that contradict each other, and the contradiction lets the model answer informational questions without ever calling `search`.

`Search requirement (MANDATORY)` says the model must run at least one search before answering anything that is not casual chit-chat, and that the first action in the turn must be the `search` tool. `Early Stop Criteria` item 1 says the model may stop as soon as "You can clearly answer the user's question with current information". Nothing anchors item 1 to a completed search, so it is satisfiable at the very start of a turn with zero tool calls. A model that believes it already knows the answer can stop before searching and still consider itself compliant.

## Evidence

All of this is visible in `lib/agents/prompts/search-mode-prompts.ts` on `main`; no runtime data is needed to see it.

1. In `getQuickModePrompt`, `Early Stop Criteria` item 1 ("You can clearly answer the user's question with current information") sits directly against `Search requirement (MANDATORY)` ("you MUST run at least one search before answering", "your FIRST action in this turn MUST be the `search` tool").
2. Item 2 of the same list is anchored: "The single search has completed, even if the available evidence is limited". Item 1 is the only entry in that list that does not reference work already performed.
3. `getAdaptiveModePrompt` has no such entry. All five of its early-stop criteria presuppose completed work (todoWrite tasks completed, multiple search angles converging, diminishing returns on further searches, strong coverage, "after 5-10 steps"). None of them can be satisfied before the first tool call. Quick's item 1 is the only stop criterion in either mode that can.
4. The `Rule precedence` block at the end of the quick prompt resolves conflicts between the search requirement and *brevity*. It says nothing about the early-stop criteria, so the prompt never resolves this particular conflict.

There is a second, downstream consequence that is not a matter of style. Citation labels are minted from search results (`assignCitationLabels` in `lib/tools/search.ts`, called from the search tool's result path). A turn that ran no search contributes no labels for that turn, so a citation the model writes anyway cannot be resolved and is dropped by `processCitations` (`lib/utils/citation.ts`). Skipping the search therefore damages grounding and citation integrity at the same time, and the two failures are not independent.

To reproduce the behavior from telemetry rather than from the prompt: for each answer, check whether a `search` span exists under the root `research` span, and split every citation metric by that flag before reading it. Blended citation rates mix searched and unsearched turns and are not interpretable.

## Proposed direction

This needs a product decision, because every option that closes the hole makes quick mode do more work, and quick mode exists to be fast and cheap.

1. **Anchor item 1 to a completed search** (for example "...with the information returned by the completed search"), matching item 2 and the adaptive prompt. Smallest, most surgical change; enforces the policy the prompt already declares.
2. **Extend `Rule precedence`** to state that the search requirement outranks the early-stop criteria, leaving item 1's wording untouched. Same effect, blunter instrument.
3. **Keep an explicit, narrow escape**: name the cases where answering with no search is intended (greetings, questions about the assistant itself, reformatting or translating text already present in the conversation) and require that no citation be emitted on such a turn.
4. **Accept the current behavior** and instead suppress the citation instruction when the turn ran no search, fixing only the citation half and leaving grounding as-is.

Options 1-3 trade cost and latency for grounding. Option 4 keeps quick cheap and accepts ungrounded answers, but stops the model from writing citations it cannot back. Choosing between those is the decision this issue is asking for.

## Acceptance criteria

- The quick prompt no longer contains a stop criterion that can be satisfied before the first tool call while a mandatory search requirement is in force, or `Rule precedence` explicitly resolves the conflict.
- The share of quick turns that answer an informational question without running any search moves decisively toward the adaptive share, judged against internal measurement rather than a number fixed here.
- On quick turns that ran no search, citation resolution stops being systematically worse than on quick turns that searched.
- The related-questions evals in CI stay inside their existing gates. Prompt edits are known to perturb instructions they do not target, so the image-spec and related-question output rates must be compared before and after.

sushanttiwari-14:
Quick mode could answer informational questions before its mandatory search, leaving answers ungrounded and citation labels unavailable. The prompt now selects one of four tool plans from whether actionable URLs are present and whether outside information is needed, and permits completion only after that plan finishes.

The policy preserves direct answers for supplied-context and self-contained requests, distinguishes incidental links from URL cards, retrieves all distinct actionable URLs before any required search, and bounds retrieval fallbacks. Empty or unusable regular output triggers one API attempt; access refusals stop retrieval. Search errors and empty results end the single allowed search without invented citations or additional calls.

Validation: 730 tests passed, 3 skipped; lint, typecheck, formatting, and production build passed locally. The final wording adjustments also passed the targeted prompt tests and formatting check. Prompt assertions protect the written contract; they do not prove model compliance. Live related-question/image-output comparisons and production search/citation measurements remain unverified because the required evaluation credentials are unavailable in this fork.

Fixes #1014

### DIFF
--- a/lib/agents/prompts/search-mode-prompts.ts
+++ b/lib/agents/prompts/search-mode-prompts.ts
@@ -47,13 +47,14 @@ You are a fast, efficient AI assistant optimized for quick responses. You have a
 
 **URL classification:**
 - A URL is actionable only when the user asks you to open, inspect, summarize, compare, or otherwise use its contents
+- When one or more URLs are the turn's only substantive content, every URL is actionable; treat the turn as an implicit request to use their contents
 - A URL included only as literal text to translate, rewrite, reformat, or reproduce in creative output is not actionable and MUST NOT be fetched
 
 **Early Stop Criteria (stop when the one applicable criterion is met):**
 1. The informational request has no actionable URLs: the single required search has completed, even if the available evidence is limited
 2. The request asks only about content in actionable URLs: a fetch attempt has completed for every actionable URL, even if some pages could not be retrieved
 3. The request asks about actionable URLs and requires broader information: fetch attempts have completed for every actionable URL AND the single required search has completed
-4. The request has no actionable URLs and needs no external information because it is limited to casual chit-chat, a question about the assistant itself, transforming user-provided text, or purely creative generation
+4. The request has no actionable URLs and needs no external information because it can be answered entirely from material already present in the conversation or attached by the user, or is limited to casual chit-chat, a question about the assistant itself, transforming user-provided text, or purely creative generation
 
 Language:
 - ALWAYS respond in the user's language.
--- a/lib/agents/prompts/search-mode-prompts.ts
+++ b/lib/agents/prompts/search-mode-prompts.ts
@@ -46,8 +46,10 @@ You are a fast, efficient AI assistant optimized for quick responses. You have a
 - After the first search result, answer immediately without another search or fetch
 
 **Early Stop Criteria (stop when ANY of these is met):**
-1. The required search has completed and returned enough information to clearly answer the user's question
+1. For informational questions without URLs, the required search has completed and returned enough information to clearly answer the user's question
 2. The single search has completed, even if the available evidence is limited
+3. The message is limited to casual chit-chat (for example, a greeting or thanks) and does not request information or advice
+4. The user provided a URL and the fetch has completed
 
 Language:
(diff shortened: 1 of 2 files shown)


miurla (approved):
LGTM 👍

Thanks for taking this on, and for making the tool plan and the terminal conditions explicit instead of patching the single criterion.

PR #1034, opened 2026-09-21
fix(search): handle paths in Tavily site-only queries

miurla:
## Problem

A search whose query is a single `site:` operator carrying a path, for example
`site:example.com/docs/guide/intro.pdf`, is sent to Tavily verbatim and the request
fails immediately with `HTTP 400: Bad Request`. The tool call errors out, so the turn
gets no results at all from that search.

This is the same failure `SITE_OPERATOR_PATTERN` already exists to prevent: Tavily
rejects a query made only of `site:` operators. The bare form
(`site:example.com`) is rewritten into `include_domains` and never reaches Tavily, but
the path-bearing form is not, so it still reproduces the rejection.

## Root cause

`lib/tools/search/providers/tavily.ts`

`SITE_OPERATOR_PATTERN = /^site:[^\s/:?#@\\]+$/i` excludes `/`, so an operand with a
path does not match. `extractSiteOnlyDomains()` returns `null` as soon as one token
fails the pattern, which sends the whole query through untouched.

The existing comment explains the exclusion as "include_domains cannot express them",
which is true of the path but does not hold for the hostname. Leaving the operand in
place keeps the part Tavily rejects, and drops the part it would have accepted.

## Fix

Site-only queries now accept operands that carry a path:

- the host part is resolved to an ASCII hostname through the existing
  `toAsciiHostname` / `normalizeDomains` filtering and goes into `include_domains`,
  exactly as the bare form does today
- the path becomes ordinary search terms (separators collapsed to spaces, trailing
  file extension dropped), so the query still points at the specific document instead
  of the domain root
- an operand whose host is not a valid domain keeps today's behaviour and the query is
  left unchanged

Out of scope, behaviour unchanged: operands carrying a port, and any query that mixes
a `site:` operator with other search terms (Tavily accepts those).

## Verification

`bun lint`, `bun typecheck`, `bun format:check`, `bun run test`, `bun run build` all
pass.

New cases in `lib/tools/search/providers/__tests__/tavily-domains.test.ts` cover the
path rewrite, a deep path with underscores and a file extension, a path-bearing
operand with an invalid host, and a mixed query. The pre-existing port and
invalid-domain cases are untouched and still pass.


miurla:
Fixed in `3168181`. The extension is now dropped only when an alphabetic suffix follows a nonempty stem, so `/.well-known` keeps its component and a version slug like `/docs/v1.2` is no longer truncated to `v1`. Added a regression test covering both.

miurla:
Fixed in the latest commit. The extension suffix now accepts digits as long as it carries at least one letter, so `.mp4` and `.7z` are dropped while a purely numeric version suffix like `/docs/v1.2` is still preserved. Regression test added for both.

PR #1038, opened 2026-09-22
fix(chat): bound cold-start PDF replay

### LINKED ISSUE #1035: PDF attachment token estimate is far below the provider's, so cold-start guards never fire on large PDFs
## Problem

The PDF attachment token estimate is far lower than what the provider actually
charges for the same file, so a thread carrying a large PDF slips past every guard
that is supposed to bound a cold-start request. The thread then replays the whole
document on each cold start, which is exactly the case `trimColdStartHistory` exists
to bound.

The gap is not a rounding error and it grows with the size of the document: the
larger the PDF, the further the estimate falls below the real weight, and it errs in
the direction that keeps the budget check from ever tripping.

## Evidence

`lib/utils/attachment-tokens.ts`

```ts
export const PDF_BYTES_PER_TOKEN = 30
```

`estimateAttachmentTokens()` divides the compressed byte size by that constant. The
result feeds three consumers:

- `lib/streaming/helpers/trim-cold-start-history.ts` (the `COLD_START_HISTORY_TOKEN_LIMIT`
  budget, default `200000`)
- `lib/streaming/helpers/cap-historical-attachments.ts`
- `lib/streaming/helpers/summarize-carried-context.ts` (the `attachmentTokens` field on
  the turn's root span)

Bytes are a poor proxy here because a provider bills a PDF by its rendered pages, not
by its compressed container, and the compression ratio of a large multi-page document
is nothing like that of a small one. The constant was last calibrated in #1025, which
moved it in the right direction but kept the single-coefficient model.

How to reproduce the measurement without production data: attach a large multi-page
PDF, read `attachmentTokens` off the turn's root span (that value is exactly
`ceil(size / PDF_BYTES_PER_TOKEN)`), and compare it with the provider-reported input
token count for the turn where the PDF first lands. The difference between that turn's
input count and the preceding turn's is the document's real token weight.

## Proposed direction

Two parts, and the second is the reason this is an issue rather than a patch.

1. Make the estimate track the real cost better than a single bytes-per-token
   constant can. Options worth weighing: a size-banded coefficient (the current value
   is calibrated for small documents and does not hold for multi-megabyte containers),
   or a page-count signal taken at upload time rather than a byte heuristic.

2. Decide what should happen once the estimate is correct, because raising it changes
   behaviour in a user-visible way. A long thread whose subject *is* the attached PDF
   would newly exceed the cold-start budget, and `trimColdStartHistory` drops whole
   turns, so the document can disappear from a conversation that is entirely about it.
   `capHistoricalAttachments` will not catch it first when the thread carries a single
   attachment, because of `MIN_WEIGHT_BLOCK_ATTACHMENTS`. So the correction needs one
   of:
   - an attachment-aware exception in the cold-start trim (keep the turn that carries
     the attachment, drop its neighbours), or
   - a placeholder path so a dropped attachment leaves a reference behind instead of
     vanishing, or
   - lowering the attachment weight cap so the attachment guard, not the history trim,
     is the one that reacts.

Correcting the coefficient on its own is not safe: it trades a cost leak for a
correctness regression.

## Acceptance criteria

- The estimate for a large multi-page PDF is within the same order of magnitude as the
  provider-reported token weight for the turn that introduces it, measured as described
  above. Small PDFs keep an estimate no worse than today.
- A cold start on a thread carrying a large PDF is bounded by the cold-start budget
  rather than replaying the whole document.
- A thread whose subject is the attached document does not lose the document from its
  replayed history. Whichever of the three options above is chosen, there is a test
  that pins this.
- No change to the guest path and no schema change.

sushanttiwari-14:
Large PDFs were underestimated using a single bytes-per-token ratio, allowing persisted chats to replay expensive documents without triggering cold-start limits.
This change:
- Uses a size-banded PDF estimate for persisted chats while keeping guest behavior unchanged.
- Replaces trimmed attachments with safe, named placeholders instead of silently losing their references.
- Preserves the current request and fitting conversation history.
- Keeps warm-request prefixes stable.
- Adds regression coverage for large PDFs, context truncation, telemetry, and attachment placeholders.
Validation
- 884 tests passed, 3 skipped
- ESLint passed
- TypeScript passed
- Prettier passed
- Production build passed
Closes #1035.


lib/streaming/helpers/trim-cold-start-history.ts
```diff
@@ -177,24 +244,65 @@ export function trimColdStartHistory(
...
+      firstTurnHasAttachments &&
+      !replaceFirstTurnAttachments
+    ) {
+      const replacementDelta = firstTurnPlaceholderTokens - tokens[0]
+      replaceFirstTurnAttachments = true
+      protectedTokens += replacementDelta
+      total += replacementDelta
+    }
+
     while (total > limit && cut < i) {
```
miurla, line 268: `protectedTokens` includes `leadingUserTokens[i]`, i.e. the user message of the turn being evaluated. When that message alone carries a large PDF, the protected set already exceeds `limit`, so this loop advances `cut` all the way to `i` (and flips `replaceFirstTurnAttachments` above) without ever bringing `total` under the limit.

Concrete case: a thread with 5+ turns, the user comes back after the cold gap and attaches a ~2 MB PDF. With `estimatePersistedAttachmentTokens` that message is ~366k on its own, above the 200k default. Every turn between the first and the current one is replaced by its attachment references, and the first turn's attachments become placeholders, yet the request is still over the limit, so the history loss buys nothing. Because `cut` is recomputed deterministically from history, those turns stay dropped on every later request in the thread.

The shape existed before, but with the previous estimator it needed a PDF of roughly 6 MB; the new curve brings the trigger down to roughly 1.5 MB. Suggestion: if `protectedTokens` (with the current turn's user message) already exceeds `limit`, skip advancing `cut` for that turn instead of cutting to `i`, and add a test for a cold current turn whose own attachment exceeds the limit.

lib/streaming/create-chat-stream-response.ts
```diff
@@ -248,10 +253,16 @@ export async function createChatStreamResponse(
...
         capHistoricalAttachments(
-          compactHistoricalMessages(coldStartHistory.messages)
+          compactHistoricalMessages(coldStartHistory.messages),
+          HISTORY_ATTACHMENT_REPLAY_LIMIT,
+          HISTORY_ATTACHMENT_TOKEN_BUDGET,
+          estimatePersistedAttachmentTokens
         )
       )
-      carriedContext = summarizeCarriedContext(messagesToConvert)
+      carriedContext = summarizeCarriedContext(
```
miurla, line 262: `carriedContext` is recorded on the root trace span and is used to compare the estimate against the provider's reported input size. Switching it to `estimatePersistedAttachmentTokens` here, while the ephemeral path keeps `estimateAttachmentTokens`, changes what `attachmentTokens` means at deploy time and makes the authenticated and guest values incomparable (a ~4.3 MB PDF goes from ~145k to ~1.15M on this path only).

I'd keep `summarizeCarriedContext` on the default estimator (it describes what was sent, not the trim's budget). If the persisted estimate is useful telemetry, it could go in a separate field.

miurla (commented):
Thanks for picking up #1035. Two things from a pass over the diff, plus a few notes that don't need to block this.

**Notes (non-blocking)**
- The persisted estimator is also passed to `capHistoricalAttachments`, so on the authenticated path it changes which historical attachments are dropped on warm turns too, not only at cold start. That may be intended, but it is broader than the PR title suggests.
- `buildAttachmentTokenEstimates` still uses the default estimator, so `shouldTruncateMessages` / `truncateMessages` and the cold-start trim now judge the same PDF with different estimates.
- The early return for `turns.length < MIN_TURNS` is unchanged, so a short thread that carries one very large PDF is still never evaluated by the trim.
- The piecewise PDF curve (1 MB at `PDF_BYTES_PER_TOKEN`, then 3 bytes/token) is a single fit. Extracted-text density varies a lot between PDFs (scanned vs text-heavy), so it would help to say in the PR what it was calibrated against.

miurla:
Thanks for this, but we're going to take a different approach: providers bill PDFs by page, and byte size can't predict page count (two ~5 MB PDFs I checked have 135 and 6 pages), so any bytes-per-token curve will be wrong in one direction or the other. We'll instead count pages at upload time, reject documents over a page limit, and use the stored page count for the estimate, which also bounds the first turn that the cold-start trim never sees. Closing this for now, and your placeholder idea will likely come back for the multi-document case in #1035.

